PR #24, opened 2023-06-21
Updates + Changes

PassiveLemon:
Some code improvements.
If the version ever changes, it will download the new binary if it doesn't already exist. Old ones will be kept but I can implement a change to remove them.

Some general changes to documentation.


PassiveLemon:
You must have updated the README while I was working on this PR so I had to undo the change to that.

PR #36, opened 2023-07-28
Docker 2.1.0

PassiveLemon:


tobychui (approved):
LGTM. Maybe we need to fix the docker versioning system to match the zoraxy release later after the beta phrase end.

PassiveLemon:
I wouldn't be opposed to having a docker release for each Zoraxy release but this may have weird backwards compatibility issues. If you think this would be a better approach, lmk.

tobychui:
@PassiveLemon you have your point. Let me think about this 👍🏻

PR #48, opened 2023-08-20
Add custom ACME server feature in backend

daluntw:
This PR add custom ACME server feature in backend, but still need frontend support (send `ca_url` parameter when POST to `/api/acme/obtainCert` and `ca` set to `custom`, should be somewhere near `src/web/snippet/acme.html#L320`, if `ca_url` not included, will use `ca` just like before)

this PR's feature will save a json in `./conf/certs/${certificateName}.json` next to the cert key for renew usage, `eab_key_id` and `eab_hmac_key` can be save in json too for future Google ACME server implement.

this also fix Buypass ACME might not be renew correctly, the reason is current implementtion in `src/mod/acme/utils.go#L57` will use Issuer organization sector as `ca.json` key, but current Buypass intermediate certificate use `Buypass AS-983163327` which might cause an issue (but seems ok to LE and ZeroSSL)

### DIFF
--- a/src/mod/acme/acme.go
+++ b/src/mod/acme/acme.go
@@ -8,7 +8,6 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"encoding/pem"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"log"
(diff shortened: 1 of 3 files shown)


tobychui:
@daluntw Thanks for the PR! 
I have briefly look at the code and it seems promising. I will be try this out in the coming weekend. 
Thank so much for your contributions!

daluntw:
I add frontend support for this feature, works perfectly on my env.

yeungalan:
We will take a look; thanks for your contribution!

tobychui (approved):
lgtm

daluntw:
I think I should add ignoreTLS option for this feature, since most selfhosted ACME server is self-signed

PR #41, opened 2023-08-11
Better docker support

Saturn745:
This PR simplifies the docker image by having it build the binary and run the built binary at runtime instead of building it at runtime like it was previously. This also includes a GitHub action to automatically build the image and publish it to Github container repo (Can be changed to Dockerhub) on each push (Can change to when a release is made if wanted).

PR #54, opened 2023-08-24
Completely rework the container

PassiveLemon:
While I would have liked to merge the PR https://github.com/tobychui/zoraxy/pull/41, I want to keep some of my more standard practices which the PR did not really have and it would just not be effective trying to get the user to format it the way I think is best.

I have tested it and it does work on Linux x64. Not sure about Arm.
Also not sure about the `on: release:` action trigger since I used Act to test the workflow file and I couldn't figure out how to trigger a release type published with a tag so I used some workaround.

I would also make a release statement that says:
```
Breaking Changes:
File structure change requires you to update the volume mount for the configurations. It should be changed to `/opt/zoraxy/config/`
The management port is no longer changeable. This is to allow for a healthcheck.

Changes:
Healthcheck was added. See breaking changes above.
Notifier was removed.
VERSION variable is no longer configurable.
```
You could probably just merge this in with main and add those statements to the 2.6.5 release notes and I'll manually upload the version to Dockerhub.


tobychui (approved):
lgtm

PR #62, opened 2023-09-03
Optimize and improve code

Xpl0itU:
This PR aims to improve code quality by removing redundant code paths and improving performance on various fields, most notably string concatenation, which is way faster when using fmt.Sprintf than with the + operator

### DIFF
--- a/src/accesslist.go
+++ b/src/accesslist.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"log"
 	"net/http"
 
 	"imuslab.com/zoraxy/mod/utils"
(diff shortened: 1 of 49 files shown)


tobychui:
Thanks for your PR. But I am seeing quite some issue with this "optimize" and "improved codes".

First of all, your code a bit more on the C language code style than Go. I have no personal preference in code style,  just I don't think it is necessary to change it (and might cause stability problem for tested & stable code). For example, replacing if else cases with switch for string is not strictly necessary (except in a few files in the acme / acme wizard module).

Next, + is not any slower than sprintf. (see [here](https://dev.to/pmalhaire/concatenate-strings-in-golang-a-quick-benchmark-4ahh)), that is why I think we can keep it the way it is right now regarding how strings are joined. 

Then, it is about the "true" "false" comparison. As the util module is what I copy from another project when I started develop Zoraxy a few years ago, it need some updates to add support like ```utils.GetBool``` or ```utils.GetInt``` to perform abstracted type conversion. Your implementation is valid but didn't make sense in the long run. 

I am sorry but I am closing this PR now. You are welcome to submit another PR if you have update anything. It would be better for you to submit a draft first before a formal PR so I can give you some comments and guide on how to contribute to this project effectively. 

Finally, thank you for your love to this project! I appreciate your wiliness to contribute and looking forward for your next improved PR!

Issue #61, opened 2023-09-03
[BUG] Let's Encrypt certs are Staging ones, so they're untrusted on browsers


Xpl0itU:
**Describe the bug**
When generating a Let's Encrypt certificate, they're generated in the staging server

**To Reproduce**
Generate a Let's Encrypt cert using Zoraxy

**Expected behavior**

**Screenshots**
![image](https://github.com/tobychui/zoraxy/assets/24777100/5c0a47e5-db2c-47b0-9a5c-7dc5e2584507)

**Host Environment (please complete the following information):**
-  Arch: x86_64
 - Device: Dell Workstation
 - OS: Debian
 - Version: 12 Bookworm (on a proxmox container)

tobychui:
@yeungalan  Can you take a look at this real quick?

tobychui:
Hi @Xpl0itU 
Although I am not in charge of the acme module, but from what I see in the source code, it is not possible to use the staging directory unless it is generated via the test case

As you can see here, the link are hardcoded and embedded into the binary
https://github.com/tobychui/zoraxy/blob/83f574e3abe208d731b018ab73a42108d15cc885/src/mod/acme/ca.go#L40

Which is referring to this link
https://github.com/tobychui/zoraxy/blob/83f574e3abe208d731b018ab73a42108d15cc885/src/mod/acme/ca.json#L3

Might I know how you build this binary and what version of Zoraxy you are using?

Xpl0itU:
I used the Zoraxy script from [here](https://tteck.github.io/Proxmox/) to setup my container

Xpl0itU:
Here's the output of zoraxy -info:
```
root@ct-zoraxy:/opt/zoraxy/src# ./zoraxy -info
{
 "Name": "Zoraxy",
 "Desc": "Dynamic Reverse Proxy Server",
 "Group": "Network",
 "IconPath": "zoraxy/img/small_icon.png",
 "Version": "2.6.6",
 "StartDir": "zoraxy/index.html",
 "SupportFW": true,
 "LaunchFWDir": "zoraxy/index.html",
 "SupportEmb": false,
 "LaunchEmb": "",
 "InitFWSize": [
  1080,
  580
 ],
 "InitEmbSize": null,
 "SupportedExt": null
}
```

tobychui:
@Xpl0itU Can I see the public key of your cert? I am guessing you didn't setup your root domain TLS certificate correctly and zoraxy is loading its internal dummy cert.

[...]

Issue #47, opened 2023-08-20
[ENHANCEMENTS] Set default SSL / TLS Provider


Morethanevil:
When adding a new subdomain proxy Zoraxy asks if a new certificate shall be created. Zoraxy then uses Let´s encrypt, but I would like Buypass SSL. So I need an option to set a default provider :)

PR #63, opened 2023-09-08
Use Let's Encrypt as default when not select CA in frontend

daluntw:
Quick fix for issue https://github.com/tobychui/zoraxy/issues/61

but should also consider issue https://github.com/tobychui/zoraxy/issues/47 in backend

### DIFF
--- a/src/web/snippet/acme.html
+++ b/src/web/snippet/acme.html
@@ -344,6 +344,8 @@
       if (ca == "Custom ACME Server") {
         ca = "custom";
         caURL = $("#caURL").val();
+      } else if (ca == "") {
+        ca = "Let's Encrypt";
       }
 
       var skipTLSValue = $("#skipTLSCheckbox")[0].checked;


tobychui:
Hi @daluntw , I have added this into the 2.6.7 branch yesterday. I guess my fix is 100% the same as yours XD
```
ca, err := utils.PostPara(r, "ca")
	if err != nil {
		log.Println("[INFO] CA not set. Using default")
		ca, caUrl = "", ""
	}

	if ca == "custom" {
		caUrl, err = utils.PostPara(r, "caURL")
		if err != nil {
			log.Println("[INFO] Custom CA set but no URL provide, Using default")
			ca, caUrl = "", ""
		}
	}

	if ca == "" {
		//default. Use Let's Encrypt
		ca = "Let's Encrypt"
	}
```

So I am closing this issue now. Thanks for your PR!

PR #65, opened 2023-09-13
Fix the out of range problem when certificate auto renew 

daluntw:
In some self-hosted acme server, the org section may not exist, and that will cause the auto renew feature error

```
2023/09/12 03:01:14 Check and renew certificates in progress
panic: runtime error: index out of range [0] with length 0

goroutine 67 [running]:
imuslab.com/zoraxy/mod/acme.ExtractIssuerName({0xc01683d500?, 0xc02e4ec240?, 0x0?})
        /opt/zoraxy/source/mod/acme/utils.go:57 +0x10d
imuslab.com/zoraxy/mod/acme.(*AutoRenewer).CheckAndRenewCertificates(0xc03e212240)
        /opt/zoraxy/source/mod/acme/autorenew.go:318 +0x385
imuslab.com/zoraxy/mod/acme.(*AutoRenewer).StartAutoRenewTicker.func1(0xc03e11c240?)
        /opt/zoraxy/source/mod/acme/autorenew.go:118 +0xbf
created by imuslab.com/zoraxy/mod/acme.(*AutoRenewer).StartAutoRenewTicker in goroutine 1
        /opt/zoraxy/source/mod/acme/autorenew.go:111 +0x107
```

and since we use the `certInfo.AcmeName`, `certInfo.AcmeUrl`, `certInfo.SkipTLS` from json file, there is no need to extract `CAName` from certificate anymore

### DIFF
--- a/src/mod/acme/autorenew.go
+++ b/src/mod/acme/autorenew.go
@@ -40,7 +40,6 @@ type AutoRenewer struct {
 type ExpiredCerts struct {
 	Domains  []string
 	Filepath string
-	CA       string
 }
 
 // Create an auto renew agent, require config filepath and auto scan & renew interval (seconds)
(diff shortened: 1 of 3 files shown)


tobychui:
Hi @daluntw,
I think the new implementation is all right, but have you considered keeping the org checking logic as fallback as some of the users out there still using an old version of zoraxy? If you changes this without keeping backward compatibility, this will be a breaking change and might cause some issues in running systems.

@yeungalan I guess you need to review this as well

daluntw:
that make sense, I add a fallback when JSON file not exist, and will create it after renew.

I also add a fallback for Buypass which org section not match the `ca.json`

tobychui (approved):
Look good to me, waiting for your review @yeungalan

yeungalan (approved):

PR #70, opened 2023-10-03
Update container (2.6.8)

PassiveLemon:
Before this gets merged, it will need an access token for the organizations packages.

I need 2 secrets:
`GHCR_USERNAME` with the value of `aroz-online` (The organizations name)
`GHCR_PASSWORD` with the value of the access token.

That should allow this workflow to publish to a package in the organizations packages.

Or you could use a personal access token with your username and publish it directly to this repository.


tobychui:
@PassiveLemon 
I have enabled public package but I don't see a place to generate those tokens
圖片

Are there any guideline / tutorial for the action you want? Thanks!

PassiveLemon:
Oh so there aren't just access keys for orgs. In that case, you would need to enable "Personal Access Tokens" then create and use one of your accounts keys for packages. If you turn "Personal access token (classic)" to Allow, it should allow access to organization resources from individual user access keys. In that case, I think you would need to use the organizations name but then your own access key.

tobychui:
> Oh so there aren't just access keys for orgs, whoops. In that case, you would need to enable "Personal Access Tokens" then create and use one of your accounts keys for packages. If you turn "Personal access token (classic)" to Allow, it should allow access to organization resources from individual user access keys. In that case, I think you would need to use the organizations name but then your own access key.

I don't think that is a good idea. Is there any alternative way for deploying to GHCR(?) without the personal access token? That feels like a security issues to my other projects. Something like webhook to an alternative git account / server and do it from there etc.

PassiveLemon:
I commented out the GHCR portion of the action so this PR currently just rearranges the action in a better way and changes the Docker port to the new management port.

tobychui:
Okey, so I just merge this into main? @PassiveLemon

PassiveLemon:
Yes, I will manually push the updated image.
If you ever change something like the port or build process of the project, let me know so I can update the container before the next release

PassiveLemon:
If you trust me enough to just push my changes without your approval then I will but I like to get your feedback

tobychui:
Hi @PassiveLemon 

I have checked your commit. Sorry I only got little to no experience in docker, but it seems you get the port wrong. The 5487 port is for static web server loopback (aka the zoraxy build-in static web server that is optionally used for proxy root), but not the port for management interface. 

I guess it would be better to open both or just keep using 8000 in your docker file.

PassiveLemon:
Oh I see. I will update that when I get home

PassiveLemon:
I reverted back to the original management port. Users that need the new port should be able to expose it themselves with a port flag

tobychui (commented):
LGTM

PR #75, opened 2023-11-04
Remove unnecessary build step. You can use multiple --tag or -t flags.

matthewbaggett:
As per title.


tobychui:
Cool! Thanks for the PR!
This seems like a docker config related PR, I will hand over the review to @PassiveLemon

PassiveLemon:
I already have these changes in an already existing PR but thanks

PR #88, opened 2023-12-06
English correction

PassiveLemon:
I just went through and fixed up some weird English. It seems to be pretty much perfect in any sub-directory within web. I didn't check each file because that would take me a long time but from the few I checked in each, they all appeared great!


tobychui (approved):
Cool, thanks for the quick fix!

PR #117, opened 2024-04-04
V3.0.1 Updates

tobychui:
+ Added regex support for redirect (slow, don't use it unless you really needs it)
+ Added new dpcore implementations for faster proxy speed
+ Added support for CF-Connecting-IP to X-Real-IP auto rewrite
+ Added better 404 page
+ Added option to bypass websocket origin check
+ Updated project homepage design
+ Fixed recursive port detection logic
+ Fixed UserAgent in resp bug
+ Updated minimum required Go version to v1.22 (Notes: Windows 7 support is dropped)


yeungalan:
LGTM approve

Issue #133, opened 2024-04-27
[BUG] Statistics are not saved, starting from zero after restart of the docker image


daniel-l:
**Describe the bug**
Statistics are not saved, starting from zero after restart of the docker image.

**To Reproduce**
1. Use Zoraxy, accessing reverse proxied services.
2. Statistics get generated (number of "Req. Today" etc.)
3. Restart Zoraxy's docker image
4. Statistics are reset and start from zero.

**Expected behavior**
Statistics get saved and are loaded after every restart of the docker image.

**Additional context**
The docker image is setup with a volume to save data/config across sessions:
```
    volumes:
      - ./zoraxy:/opt/zoraxy/config/
```
Configs like the set up proxy hosts and custom access lists get saved there; stats not so much.

PassiveLemon:
I believe this would depend on how Zoraxy itself caches statistics. If it caches to a specific location, you could create a volume for it. I'll have to wait and see what Toby says to know for sure

tobychui:
Cannot reproduce this bug while running Zoraxy natively without docker.

@PassiveLemon For your information, the statistic is stored in memory and only write to `sys.db` on either the daily ticker ticks (everyday at midnight), or when the application receives an `os.Interrupt` or  `syscall.SIGTERM` signal (e.g. Ctrl +C on a running terminal window). You might want to check if the application is shutdown correctly or if the sys.db is persistence after restart.

daniel-l:
The file sys.db already existed yesterday; I just checked if the modified time updated since:
```
$ ls -al sys.db
-rw------- 1 root root **122880** 28. Apr **02:00** sys.db
```
The file also grew a bit since yesterday.

Also, I just shut down and restarted the container - yesterday's stats were indeed saved. Everything since (today's stats) were thrown away. File info after the restart:

```
$ ls -al sys.db
-rw------- 1 root root **122880** 28. Apr 11:43 sys.db
```

tobychui:
Ok, this surely seems like a "docker not gracefully closing zoraxy" issue. I will let @PassiveLemon  take over from here. Thanks for the bug report!

PassiveLemon:
That would make sense, there is no graceful shutdown implemented in the container. Docker sends out a SIGTERM to the running process (shell in this case) but it doesn't respond to it so, after the grace period, it SIGKILLs everything which would result in an improper save.

CorneliusCornbread:
[...]

PR #135, opened 2024-04-28
Fix: Graceful container shutdown

PassiveLemon:
Fixes https://github.com/tobychui/zoraxy/issues/133

Slight refactor to remove the useless entrypoint script and run Zoraxy as PID 1 instead so it receives the container SIGTERM and can properly shutdown.


tobychui (approved):
LGTM

PR #152, opened 2024-05-15
V3.0.4

tobychui:
This PR tidied up the previous PR related to DNS challenge and rewrite the section that require the use of os.Env variables. This PR might also create some changes to the port uses and ACME module which require testing before merge.

- Optimized DNS challenge implementation
- Removed dependencies on environment variable write and keep all data contained
- Fixed panic on loading certificate generated by Zoraxy v2
- Added automatic form generator for DNS challenge / providers
- Added CA name default value
- Added code generator for acmedns module (storing the DNS challenge provider contents extracted from lego)
- Fixed ACME snippet "Obtain Certificate" concurrent issues in save EAB and DNS credentials


tobychui:
@yeungalan has tested in on his VPS instance and everything seems all right. If there are no further issue need to address regarding DNS challenge support in ACME, this will be merged by next week. 👍🏻

PR #168, opened 2024-05-23
Use correct Host HTTP header

### LINKED ISSUE #164: [BUG] Generally buggy behaviour when proxying Nextcloud
So I'm attempting to proxy my nextcloud instance, I've got the trusted IPs setup in nextcloud and have had this nextcloud instance proxied successfully via NPM in the past. However, it seems like Nextcloud really doesn't like being proxied via Zoraxy. Looking at the browser console, it seems like it's not properly proxying resource requests to be done via the domain, rather trying to directly access the resources via either the docker IP or the local IP address.

image

**To Reproduce**
Nextcloud proxy:
image
Changing to the local IP address instead of the docker one will simply result in it working, but only VIA LAN connections.

**Expected behavior**
Nextcloud should load as normal.

**Browser (if it is a bug appears on the UI section of the system):**
 - OS: Fedora (tested non LAN connections via phone data on Android)
 - Browser: Firefox
 - Version: 126.0

**Host Environment (please complete the following information):**
-  Arch: x86
 - Device: Ryzen 3600x system
 - OS: Unraid
 - Version: 6.12.10

**Additional context**
I cannot reproduce this behaviour with my Crafty or Dozzle instance, they work just fine.

nettybun:
See https://github.com/tobychui/zoraxy/issues/164#issuecomment-2127838406.

Please advise if this is safe to do. I didn't understand the original comment, it is likely that you know better if this is a dangerous change. It does fix Nextcloud though and would close #164. Thanks!

### DIFF
--- a/src/mod/dynamicproxy/dpcore/dpcore.go
+++ b/src/mod/dynamicproxy/dpcore/dpcore.go
@@ -346,11 +346,8 @@ func (p *ReverseProxy) ProxyHTTP(rw http.ResponseWriter, req *http.Request, rrr 
 	p.Director(outreq)
 	outreq.Close = false
 
-	if !rrr.UseTLS {
-		//This seems to be routing to external sites
-		//Do not keep the original host
-		outreq.Host = rrr.OriginalHost
-	}
+	// Always use the original host, see issue #164
+	outreq.Host = rrr.OriginalHost
 
 	// We may modify the header (shallow copied above), so we only copy it.
 	outreq.Header = make(http.Header)


tobychui:
Hi @nettybun, thanks for the pull request!

Regarding the comments, usually speaking, within a trusted network like your home's LAN, you do not need TLS for connecting a service. In the original Zoraxy design, when you selected "target site require TLS connection", my assumption is that you are connecting to an external site e.g. `github.com` and with the header hidden, you can reverse proxy a site that is not design to be proxied. This is really helpful when you are trying to embed external resources on your site where the original site do not allow iframe. So with that Host header rewrite bypass, you can do something like this: 

圖片

TL;DR, This is a safe PR and should be fine merging this into the main branch. It will be better to add a check to see if it is an external site (proxy target's is a domain names and not ending with e.g. `.local` or `.internal`), but I guess I can add it later if needed.

tobychui:
Tested working in my portainer docker setup. Good works!

Issue #161, opened 2024-05-20
Configuration for OVH DNS Challenge


tadikwa:
I have some problems to configure OVH DNS Challenge

I have this field with the ACME TOOL : FieldType, SubDomain, Target and Zone
![2024-05-20_15h00_57](https://github.com/tobychui/zoraxy/assets/6054289/333d8666-1a00-425f-9bea-325a70dc5d2e)

But, normally with Nginx Proxy Manager (for example), i need to use this variables : application_key, application_secret, consumer_key and endpoint
![2024-05-20_15h01_25](https://github.com/tobychui/zoraxy/assets/6054289/5ff629b9-70d8-42fa-953b-b085bf9014c2)
![2024-05-20_15h09_09](https://github.com/tobychui/zoraxy/assets/6054289/48201bb2-fb5f-4c70-8ed8-1603c0cbb30c)

It is a bug of config parser or i didn't understand how to configure ?

tobychui:
This seems like a bug in the config parser. Let me check it in a moment.

tadikwa:
Ah yes ok I see.
However, I am running the version in Docker.
Can I also make the modification on the online container?

tobychui:
@tadikwa I guess you can. I am not experience with docker but in theory this should work.

1. Cross compile Zoraxy on your computer for your server's CPU architecture & OS, e.g. `GOOS=linux GOARCH=amd64 go build -o zoraxy-linux-amd64`
2. ssh into your container, stop zoraxy
3. Transfer the new binary from your computer to the container and overwrite the old version of zoraxy binary executable
4. Restart zoraxy service

tadikwa:
ok, this is out of my scope.

I prefer to wait for a new build and update my docker image ;)

tobychui:
@tadikwa As this is only a front-end issue, you can hack it by using the F12 developer console and enter this.

```javascript
$.ajax({
        url: "/api/acme/autoRenew/setDNS",
        method: "POST",
        data: {
          filename: your_domain_with_*_replaced_by_underscore,
          dnsProvider: "ovh",
          dnsCredentials: JSON.stringify({
               "ApplicationKey":"",
               "ApplicationSecret":"",
               "ConsumerKey":""
          }),
        }
});
```

PassiveLemon:
You could clone the repo, change the line, and then rebuild the container

barto95100:
Maybe upgrade the official container zoraxy to have this change ? most simply for all user :)

[...]

Issue #149, opened 2024-05-09
[BUG] Downstream proxy server is online with misconfigured settings


mflagler:
**Describe the bug**
Downstream Inductive Automation Server is working, but uptime monitor shows "Downstream proxy server is online with misconfigured settings"

**To Reproduce**

1. Point HTTP Proxy to self-hosted Inductive Automation Gateway Server downloaded and installed from here: [https://inductiveautomation.com/downloads/archive/8.1.39](https://inductiveautomation.com/downloads/archive/8.1.39)
2. Uptime monitor shows misconfiguration

**Expected behavior**
Uptime monitor shows healthy running server

**Screenshots**
![image](https://github.com/tobychui/zoraxy/assets/17576920/49a0aafc-6e16-4d83-be8c-f4143f4c4c07)

**Host Environment (please complete the following information):**
-  Arch: Windows x64
 - Device: Windows VM for Testing
 - OS: Windows 11
 - Version: 23H2

**Additional context**
Attached Wireshark capture with first part doing an edit/save of the HTTP proxy to force it to attempt to re-check the status/uptime. Starting at packet 43 I connected to the same URL (IP and port) as configured directly using Microsoft Edge to show how the browser packets look.

Ignition can be downloaded if wanted for testing by right-clicking the download and doing a save-as to bypass the registration form. A default install with standard edition and doing the initial setup (creating an admin user/password and specifying the ports to use) is all that's needed.

[Zoraxy Uptime Monitor Capture.zip](https://github.com/tobychui/zoraxy/files/15264379/Zoraxy.Uptime.Monitor.Capture.zip)

tobychui:
I have analysis your capture and found out that the capture didn't actually include the uptime monitor request. But anyway, from your screenshot, I think the downstream server is returning a status code in the 5xx range with the misconfigured as a badly chosen error message. I have updated the uptime monitor with better description and UI to reduce users confusion on the uptime status of their servers.

![圖片](https://github.com/tobychui/zoraxy/assets/24617523/1b3730e2-38ce-4a63-b27d-6be4ea32e4a1)

Will be closing this after next release.

mflagler:
On my packet capture, packet 24 is the request from Zoraxy and in 25 it gets a 302 Found, but then Zoraxy sends on packet 28 a continuation which then triggers a response on packet 30 of a 400 Bad Request.

[...]

Issue #121, opened 2024-04-06
[BUG] Uptime not accurate


RickoT:
**Describe the bug**
The uptime indicator is inaccurate if there is a redirect to a namespace (for example airsonic server shows "Target online but not accessible" because it returns a 403 before it redirects to the login screen)

**To Reproduce**
1. Create an airsonic docker container
2. configure for reverse proxy
3. Create a reverse proxy to the docker container
4. Watch monitor

**Expected behavior**
Check should follow redirects to make sure the site is up before indicating it is down

**Host Environment (please complete the following information):**
-  Arch: AMD64
 - Device: Docker
 - OS: Debian 12.5

**Additional context**
It would be nice to be able to adjust the health check to a specific URL if necessary

tobychui:
403 Forbidden means not accessible, and if a redirect is needed, the back end should response with 3xx. If there is a redirection done with 403, it means it is done using HTML tags or JavaScript located in the body section of the requested page.

Zoraxy only request for header in uptime to reduce resources usage, so the uptime said "Target online but not accessible" is a very accurate description. The issue here is the back end server where you are reverse proxying (airsonic server I guess?) do not follow standards HTTP guideline on how to correctly handle redirects. 

There is nothing I can do except telling you that these error message and color blocks were from this html file. If you don't like seeing orange blocks, you can change them to other colors. 

https://github.com/tobychui/zoraxy/blob/2cdd5654ed898651482883ed08d9d9b79da095ef/src/web/components/uptime.html#L166

PR #174, opened 2024-05-26
V3.0.5

tobychui:
### Change Log
- Optimized uptime monitor error message #121 
- Optimized detection logic for internal proxy target and header rewrite condition for `HTTP_HOST` #164 
- Fixed ovh DNS challenge provider form generator bug #161 
- Added permission policy module (not enabled)
- Added single-use cookiejar to uptime monitor request client to handle cookie issues on some poorly written back-end server #149


yeungalan (commented):
lgtm

PR #202, opened 2024-06-15
add docker containers list to set rules

7brend7:
## Add a popup window with avaliable docker containers and ability to populate name, ip and port into rules field

https://github.com/tobychui/zoraxy/assets/2050897/9afd8dda-1cee-46e1-8ba8-70d83fae4ee4

### Overview
I use an Rpi4 as a home server, installing each service as a separate Docker container (including Zoraxy). When configuring proxy rules, I need to set the IP and port for each service, requiring me to check the port each time. To simplify the process of adding proxy rules for Docker containers, I've added a simple popup window.

### Requirements
- To get a list of Docker containers, we need to map `/var/run/docker.sock` into the Zoraxy container.
- Because Zoraxy runs in a container, I need to set the host IP (Rpi4) as the target address to make it work. Alternatively, `host.docker.internal` can be used.
browser -> Zoraxy -> Rpi4 -> service

### Flow
When the popup opens:
- Get the current list of proxies.
- Get Docker networks (fetch the `bridge` network to get the host's IP (172.17.0.1)).
- Get the list of containers.
When the `add` button is clicked:
- Close the popup and populate the fields: name = container's name, target = address:port

### DIFF
--- a/src/api.go
+++ b/src/api.go
@@ -222,6 +222,8 @@ func initAPIs() {
 	authRouter.HandleFunc("/api/info/pprof", pprof.Index)
 
 	//If you got APIs to add, append them here
+	// get available docker containers
+	authRouter.HandleFunc("/api/docker/containers", handleDockerContainersList)
 }
 
 // Function to renders Auth related APIs
--- a/src/docker.go
+++ b/src/docker.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/client"
+	"imuslab.com/zoraxy/mod/utils"
+)
+
+func handleDockerContainersList(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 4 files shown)


tobychui:
Cool! Might I know if this only works for Zoraxy running inside a docker container, but not in Zoraxy running on a host that also have docker installed? If yes, I guess I might need to add some checking logic and this will need to be changed to merge into the v3.0.7 branch instead of main.

7brend7:
@tobychui, it will also work if Zoraxy is installed on the host. All we need is access to the `docker.sock` file to get a list of containers.

The issue is deciding which IP to display in that case. Using the host's internal IP (172.17.0.1) can work, but setting it to localhost is better. Therefore, we need a dropdown with Docker network values.

Additionally, Docker might be missing, so there's no reason to show the button and popup without an extra check. I suggest creating another branch for this feature, merging the current PR, and continuing the work there.

tobychui (approved):
lgtm

ExXxtr3me:
Nice addition. 👍 
Would it be possible to add the dockername instead of the ip adress in the target field? The IPs change sometimes on my unRaid Server.

PR #210, opened 2024-06-20
Update CHANGELOG.md

Morethanevil:
Updated changelog, providing an example configuration for docker


tobychui (approved):

PassiveLemon:
I think you should revert this for now, those environment variables are not set like that and will not do anything

tobychui:
> I think you should revert this for now, those environment variables are not set like that and will not do anything

The example has been removed from the file in the follow up commit. Checkout the latest version [here](https://github.com/tobychui/zoraxy/blob/main/CHANGELOG.md#v307-20-jun-2024)

PR #246, opened 2024-07-16
Update 3.0.9

tobychui:
This is a quick fix for the docker image and with a few fine tuning features
- Fixed docker file bug
- Added certificate download
- Updated netcup timeout value
- Updated geoip db
- Removed debug print from log viewer
- Upgraded netstat log printing to new log formatter
- Improved updater implementation


tobychui:
@PassiveLemon can you have a quick check if this works with the new docker file?

tobychui:
Ok it seems it is working. I will just merge this to avoid further issues poping up regarding the docker image not working. Please let me know if there are further issues that need a new release to fix. 

Thank you for everyone involved!

Issue #248, opened 2024-07-16
[HELP]How Can I setup my time zone?


klisza1993:
As in title, is there any way to set it? ;)

tobychui:
@klisza1993 Timezone for what?

klisza1993:
For container. Cuz ive got a few hours shifted

mbah-dukun:
you can see example from my docker compose in [ here ](https://github.com/tobychui/zoraxy/issues/241#issuecomment-2228072937)

Morethanevil:
> you can see example from my docker compose in [ here ](https://github.com/tobychui/zoraxy/issues/241#issuecomment-2228072937)

Will add the timezone to my example file in the wiki. Did not know this was a problem

PassiveLemon:
> please add in this [README.md](https://github.com/tobychui/zoraxy/tree/8e0e9531e71df2fc129c602be05ac6588f3cce0f/docker)

Will do

PassiveLemon:
Closing this as it appears to be solved

PR #251, opened 2024-07-16
Doc: Document on how to use host time in the container

PassiveLemon:
Addresses https://github.com/tobychui/zoraxy/issues/248

Also included a change to the Dockerfile to target the latest Alpine version


tobychui (commented):
LGTM

PR #258, opened 2024-07-22
weighted random upstream

bouroo:
## upstreams
- change from `sort.Slice` to newer `slices.SortFunc`
  - advice from https://pkg.go.dev/sort#Slice
> Note: in many situations, the newer [slices.SortFunc](https://pkg.go.dev/slices#SortFunc) function is more ergonomic and runs faster.

## utils
- reduce string transform


src/upstreams.go
```diff
@@ -1,9 +1,10 @@
 package main
 
 import (
+	"cmp"
```
tobychui: cmp package was introduced in go v1.21, but the last version that support Windows 7 (The NT6.1 build) was Go 1.20. Hence, this cannot be used in Zoraxy v3 (maybe you need to wait until Zoraxy v4 after we dropped windows 7 support before cmp package can be used)

src/upstreams.go
```diff
@@ -33,19 +34,22 @@ func ReverseProxyUpstreamList(w http.ResponseWriter, r *http.Request) {
 
 	activeUpstreams := targetEndpoint.ActiveOrigins
 	inactiveUpstreams := targetEndpoint.InactiveOrigins
-	// Sort the upstreams slice by weight, then by origin domain alphabetically
-	sort.Slice(activeUpstreams, func(i, j int) bool {
-		if activeUpstreams[i].Weight != activeUpstreams[j].Weight {
-			return activeUpstreams[i].Weight > activeUpstreams[j].Weight
+	slices.SortFunc(activeUpstreams, func(i, j *loadbalance.Upstream) int {
```
tobychui: This has a similar problem as well, the slices package was not introduce in Go 1.20 so shall not be used in Zoraxy v3.

tobychui (changes requested):
Hi @bouroo, thanks for the PR!

I see you try to optimize a few things in the upstream and utils package, but seems you have forgotten that we still support Windows 7 (NT6.1) in Zoraxy v3 in which part of the package you use was not compatible with the backward compatibility support. 

I do agree with your optimization in the utilities function and I really think it worth a merge, but you included both changes in 1 PR. If you want this to get merged, I guess you could update the upstreams sort function to a more compatible design for Go 1.20 or split the changes in upstreams into a separated PR.

src/upstreams.go
bouroo: Ok, I'll taking care of compatible with go 1.20 and will let you know when it's done.

ps. I'm working on the weighted random upstreams function. It should return immediately if there is only one upstream or if the upstream list is empty. Should this be included in this PR, or should I open a new PR?

tobychui: You can have both of them in one PR, I can review them at the same time as soon as you have well documented comments that shows what does what in the code.

src/upstreams.go
bouroo: I've changed back to manual compare function

src/upstreams.go
bouroo: @tobychui Ok, ready for review. [8a8ec1c](https://github.com/tobychui/zoraxy/pull/258/commits/8a8ec1cb0be387121ba13b0ac1b450952d80bd53)

tobychui (approved):

PR #259, opened 2024-07-22
Update README.md

NeoMod:
Updated the Docker Compose file to reflect the proper way of handling environment variables, especially since v. 3.0.9 introduces a breaking change compared to the composing file's previous example.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -68,6 +68,6 @@ services:
       - /var/run/docker.sock:/var/run/docker.sock
       - /etc/localtime:/etc/localtime
     environment:
-      PORT: "8005"
-      FASTGEOIP: "true"
+      - PORT=8005
+      - FASTGEOIP=true
 ```


PassiveLemon:
the `PORT: "8005"` format is [equivalent](https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-environment-attribute) to the `- PORT=8005` and I much prefer the former.

NeoMod:
@PassiveLemon I wouldn't have suggested it If I hadn't noticed that using the former format lead to the loss of functionality of the container. 
Anyway, more tests are underway since I believe I found a bug with that particular variable exposed (docker compose + portainer)

NeoMod:
@PassiveLemon Also, that would not explain the presence of this note in the docker-compose file in the repository:

image

Why making a clear point about not using the colon in the port flag, if it doesn't cause an issue? I think this should be re-evaluated.

PassiveLemon:
Thank you for catching that, I noted that as a general example for how it should look by not including the colon but it would make sense for it to be accurate, I'll update that

> Why making a clear point about not using the colon in the port flag, if it doesn't cause an issue?

Including a colon does cause issues because of how Zoraxy uses the port argument internally. When the colon is included, the healthcheck command would be `nc -vz 127.0.0.1 :8000` which is incorrect. However, when the PORT variable is passed to Zoraxy as a CLI argument, the colon is prepended: `-port=:${PORT}`

> I wouldn't have suggested it If I hadn't noticed that using the former format lead to the loss of functionality of the container.

I'm not exactly sure how the format of the environment attribute would lead to a loss of functionality. What functionality is lost? Please do update us about that

PassiveLemon:
Since this is now outdated, I'm going to close it

PR #280, opened 2024-08-03
Fix existings containers list in docker popup

7brend7:
The `api/proxy/list?type=host` endpoint doest return `Domain` value anymore.
Get ip an port from `ActiveOrigins.OriginIpOrDomain` instead.


tobychui:
Hey @7brend7, thanks for the PR!
As I am currently on vacation, I will check it out later.

tobychui (approved):
LGTM

PR #295, opened 2024-08-29
Add ZeroTier to Docker container

PassiveLemon:
When the env variable `ZTAUTH` is set, ZeroTier is started in daemon mode, otherwise it is not run at all. With ZeroTier running, Zoraxy seems to get the GAN network controller ID and I can create a network.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+if [ "$ZEROTIER" = "true" ]; then
+  echo "Starting ZeroTier daemon..."
+  zerotier-one -d
+fi
+
+echo "Starting Zoraxy..."
+exec zoraxy \
+  -autorenew="$AUTORENEW" \
+  -cfgupgrade="$CFGUPGRADE" \
+  -docker="$DOCKER" \
+  -earlyrenew="$EARLYRENEW" \
+  -fastgeoip="$FASTGEOIP" \
+  -mdns="$MDNS" \
+  -mdnsname="$MDNSNAME" \
+  -noauth="$NOAUTH" \
+  -port=:"$PORT" \
+  -sshlb="$SSHLB" \
+  -version="$VERSION" \
+  -webfm="$WEBFM" \
+  -webroot="$WEBROOT" \
+  -ztauth="$ZTAUTH" \
+  -ztport="$ZTPORT"
+
--- a/README.md
+++ b/README.md
@@ -4,7 +4,6 @@
 
 A general purpose HTTP reverse proxy and forwarding tool. Now written in Go!
 
-
 ### Features
 
 - Simple to use interface with detail in-system instructions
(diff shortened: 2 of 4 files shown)


docker/entrypoint.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+if [ -n "$ZTAUTH" ] && [ "$ZTAUTH" != "''" ]; then
```
tobychui: Hi @PassiveLemon , I am a bit concern about this implementation here. That `authtoken.secret` should be automatically generated by zerotier-one which is also used to generate the node's UUID within the network. I am not sure what is the effect of this implementation (writing to `authtoken.secret` into zerotier system file, instead of letting zerotier generate it automatically) will cause any runtime issue with multiple Zoraxy docker nodes in the same network.

tobychui (commented):
Thanks for the PR! Nice work as always :)

I have raised a concern (see comment) regarding the logic for writing the ZTAUTH parameter directly into the zerotier system file. Would it be possible that an alternative implementation (maybe something like a new variable maybe?) is used so this potential issue can be avoided in the first place? 
Many thanks!!!

P.S. I see that you are merging into the main branch. I assume that you did that assuming I will merge this before the v3.1.1 branch right? If no, please let me know.

docker/entrypoint.sh
PassiveLemon: I must not be understanding how the GAN controller is supposed to work with ZeroTier then. I placed it at that location because it's mentioned in README [Global Area Network Controller](https://github.com/tobychui/zoraxy?tab=readme-ov-file#global-area-network-controller).

Upon looking more closely at it, I think I do understand now. ZeroTier creates the key at the location and Zoraxy should automatically pick it up. The variable is only there in the event that Zoraxy can't find the key. Is that right?

PassiveLemon:
> P.S. I see that you are merging into the main branch. I assume that you did that assuming I will merge this before the v3.1.1 branch right? If no, please let me know.

Correct, I didn't want to push directly to your PR

docker/entrypoint.sh
tobychui: @PassiveLemon Yes, you understand it correctly. The variable is there just in case Zoraxy do not have permission to read the `authtoken.secret` file but the user still want to use ZeroTier function via API calls to the zerotierd daemon.

PassiveLemon: Okay, I will make the appropriate changes

PassiveLemon: I have the auto-generated part done, it picks up the token and allows network creation. However I was testing to see what happens if I do provide a ZTAUTH token, and when I do, I get a permission error. Not sure what the expected behavior is.

PassiveLemon:
Updated, however I have an enhancement at https://github.com/tobychui/zoraxy/issues/296 that I think should be looked at

PassiveLemon:
ZeroTier token is automatically generated and retrieved upon starting the container. Also added documentation for mounting the ZeroTier data directory so it isn't lost.

I'm trying to test what ZTAUTH itself does. If I manually make the authtoken.secret file unreadable, and pass in the same key that was in the file, I get "Access Denied". As far as I know, the point of that argument is to resolve a situation like this, am I missing something?

tobychui:
>  As far as I know, the point of that argument is to resolve a situation like this, am I missing something?

You sure you only disable access to that file with Zoraxy's user group but not zerotierd's user group? If zerotier cannot read that file as well, even Zoraxy can establish connection to zerotierd, zerotierd will return error message in that regard.

PassiveLemon:
I didn't even think about the fact that ZeroTier would also need to read the file haha. Both are run under the same user so the permissions affected them both equally. However, the only time this issue would really ever happen is in a situation we can't control

tobychui (approved):

Issue #303, opened 2024-09-13
DockerHub image not updated


elBelgg:
Latest version in DockerHub is 3.1.0 instead of 3.1.1

PassiveLemon:
Yep we're working on it. An issue occurs with ZeroTier when building which causes the GitHub action to fail.

PassiveLemon:
Oops didnt mean to close it

elBelgg:
Aah okey, I didnt know. Good luck with it :)

tobychui:
@PassiveLemon Maybe we could include the older zerotier version in the next release. Currently I have no clue on how to fix it (nor why the issue only happens with newer version of zerotier) 😥 .

PassiveLemon:
Sure, I could just have it build a version that works

Issue #299, opened 2024-09-04
[BUG] ZeroTier does not work in 3.1.1


PassiveLemon:
I probably should have seen this coming but `zerotier-one`(1.10.2) was removed in Alpine 3.17, but currently we are using Alpine 3.20. In the image I download it from 3.16 and install it but the GitHub action fails at this step due to untrusted signatures, so the image never finishes building and isn't pushed to Dockerhub. I don't know why it works locally but that's a different issue.

I am now trying to refactor the Dockerfile to build ZT from source and I've encountered a strange issue. Zoraxy appears to get the controller ID but when I press the button to create a new network, nothing happens and I don't see any error messages in the Zoraxy logs.

I was afraid the problem had to do with Alpine's musl, so I installed `libc6-compat` and  `libstdc++` to provide compatibility, otherwise ZT fails with unrecognized symbol errors. This did not solve it. Currently I attempted to mitigate this by building everything on Debian derivatives but I actually have the same issue.

Was any of the ZeroTier stuff modified in 3.1.1? Does Zoraxy currently support ZeroTier 1.14.0?

tobychui:
I tested the latest zerotier in a Debian VM and I notice it is kinda broken. 

- I can list the zerotier node ID, but cant create / list any network 
- zerotier-cli is working and curl access of the API endpoint is working
- Zoraxy v3.1.1 works on zerotier v1.8.3 and v1.10.1

I will investigate further to see what might be the issue here.

PassiveLemon:
After building like 6 versions of ZeroTier, it seems like the issue starts at version 1.12.0, which just so happens to have an ungodly changelog, so something probably changed in that version.

tobychui:
Closing this as zerotier has already moved to plugin and is not part of the core.

PR #305, opened 2024-09-14
Fix: Build older version of ZeroTier

PassiveLemon:
https://github.com/tobychui/zoraxy/issues/299 https://github.com/tobychui/zoraxy/issues/303

Anything from 1.12.0+ just doesn't work on Zoraxy so we build from a known good version. We also base the image on Ubuntu instead because ZeroTierOne is natively supported.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker.io/golang:alpine AS build
+FROM docker.io/golang:alpine AS build-zoraxy
 
 RUN mkdir -p /opt/zoraxy/source/ &&\
     mkdir -p /usr/local/bin/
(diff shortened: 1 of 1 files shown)


tobychui (approved):

tobychui:
@PassiveLemon Thank for the PR! 

I am thinking should I release this as v3.1.1r2 or you think waiting for the next release (v3.1.2) will be a better arrangement? As I am working on the SSO feature, I am not sure how long would it take for me to finish v3.1.2 and the current demands for v3.1.1 support on docker.

PassiveLemon:
Because ZeroTier fails to build in the GitHub action, the Docker image wasn't actually pushed so I think a 3.1.1r2 would be appropriate. That way users of the Docker image can use 3.1.1

PassiveLemon:
Thank god it built, I don't know why it took 4 and a half hours but it succeeded

PR #362, opened 2024-10-29
Make redirection URL links absolute links

rvenutolo:
This changes the redirection URL links from relative links (ex: `zoraxy.mysite.net/my-redirect.mysite.net`) to absolute links (ex: `my-redirect.mysite.net`).

I think this was the intent for the redirection page. If not, please ignore.

PS - thanks for a great piece of software.

### DIFF
--- a/src/web/components/redirection.html
+++ b/src/web/components/redirection.html
@@ -164,7 +164,7 @@
         $.get("/api/redirect/list", function(data){
             data.forEach(function(entry){
                 $("#redirectionRuleList").append(`<tr>
-                    <td><a href="${entry.RedirectURL}" target="_blank">${entry.RedirectURL}</a></td>
+                    <td><a href="//${entry.RedirectURL}" target="_blank">${entry.RedirectURL}</a></td>
                     <td>${entry.TargetURL}</td>
                     <td>${entry.ForwardChildpath?"<i class='ui green checkmark icon'></i>":"<i class='ui red remove icon'></i>"}</td>
                     <td>${entry.StatusCode==307?"Temporary Redirect (307)":"Moved Permanently (301)"}</td>


tobychui:
@rvenutolo Thanks for the PR and yes. the redirection UI is supposing to be absolute or relative with the original (sub)domain prepended. I think this is something that will need to be fixed. 

But there just one problem, your fix doesn't completely fixed the issue. Your PR only tackled the issue with absolute url but not relative path. (e.g. someone might want to redirect a user from `example.com/redirect` to `example.com/destination`) I guess you need to check if the redirection target contains a TLD or otherwise, keep it relative.

rvenutolo:
Just to be clear and make sure we're on the same page, let me be a bit clearer and detailed.

Let's say my Zoraxy instance's URL is `zoraxy.example.com`. I have entries to redirect from `redirect.example.com` to `target.example.com` and from `example.com/redirect` to `example.com/target`.

On the redirect web page (at `https://zoraxy.example.com/web/#redirectset`) all the href links in the "Redirection URL" column are all relative to the Zoraxy URL. In this example, the links are `https://zoraxy.example.com/web/redirect.example.com` and `https://zoraxy.example.com/web/example.com/redirect`. 

I assume the links should instead be `https://redirect.example.com` and `https://example.com/redirect`. If that's not the intention and I'm not understanding a different use-case than my own, that's fine. I just want to be clear.

Regardless, I love Zoraxy. It hits that perfect middle ground of doing everything I want it to without being too complicated or confusing. Thanks again.

tobychui:
@rvenutolo Ok, you have your point.

What I mean is that as the redirect rules allow regular expression, there might be chances that the redirection is relative to the subdomains of the request. In such case, when the user click on the UI, the redirection URL shd be rewritten to the subdomain that provides such redirection.

Now you mention about that, maybe removing the "clickable" link on redirection targets (or sources) that is does not start with `http://` or `https://` will be a better implementation?

Issue #330, opened 2024-10-21
[ENHANCEMENTS] Search bar and all Docker Containers


EasyArt:
On the HTTP Proxy Page there should be a search bar ctrl+f filtering the entrys (if you have many entrys its hard to find a specific one)

The Pick from Docker Containers button only shows containers with published ports. I rarely publish ports and zoraxy is member of the same network so zoraxy can connec to the internal ports of the containers it should be possible to select them. (Maybe a checkbox to activate all containers)

tobychui:
Hi @EasyArt 

I feel bad for not replying you with this feature request, but I just wanna to let you know that there are no docker developer for this project (all docker features are contributed by independent users). If you want this feature in Zoraxy, you have to wait for a contributor. Of course, if you wanna give it a try, PRs are always welcomed :)

PR #391, opened 2024-11-11
list containers with unexposed ports

eyerrock:
This pull request addresses issue #330 by adding a checkbox to the Docker container list, allowing users to display containers with unexposed ports.

When adding a container with unexposed ports as a destination, the container's name is used as the address, letting Docker handle the resolution. Please note that this requires both Zoraxy and the target container to be on the same network.

Screenshot:
image
- both containers `nginx-same-network` and `nginx-other-network` do not expose their ports
- `nginx-same-network` shares a network with Zoraxy and is therefore reachable
- `nginx-other-network` does not share a network with Zoraxy and will be unreachable.

An alternative approach could be to read Zoraxy's container ID from the filesystem and check which other containers share a network with it. However, this method has not been implemented in this pull request.

Please let me know if you have any questions or concerns.

### DIFF
--- a/src/web/snippet/dockerContainersList.html
+++ b/src/web/snippet/dockerContainersList.html
@@ -2,7 +2,7 @@
 <html>
   <head>
     <!-- Notes: This should be open in its original path-->
-    <meta charset="utf-8">
+    <meta charset="utf-8" />
     <link rel="stylesheet" href="../script/semantic/semantic.min.css" />
     <script src="../script/jquery-3.6.0.min.js"></script>
     <script src="../script/semantic/semantic.min.js"></script>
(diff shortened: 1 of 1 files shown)


tobychui:
Hi @eyerrock, thanks for your PR!

I will ping our docker maintainer @PassiveLemon to test this and see if there are any issues needed to be fixed.
If everything works, please let me know so I can merge this PR. Thank you!

PassiveLemon:
I haven't been able to test the actual proxying but I cloned the changes and looked around for a bit, it looks good! I'll try to set up some proxies later

PassiveLemon (approved):

tobychui (approved):

Issue #388, opened 2024-11-11
[BUG] ZeroTier not fully disabled


LeonardoIz:
**Describe the bug**
Even if zerotier is disabled, it is reported as Link Error, and it also automatically creates a volume in the container

**To Reproduce**
Create a container with the latest version and manually disable ZeroTier in the environment variables

**Expected behavior**
Appear as disabled, and do not create the volume automatically

**Screenshots**
![image](https://github.com/user-attachments/assets/7fe07dc5-a86d-4d5d-a6cd-e17ee8f4049a)
![image](https://github.com/user-attachments/assets/6c7ae9c3-5ed4-4da1-a712-ae79576d6730)
![image](https://github.com/user-attachments/assets/6b568c51-6999-4647-b005-3c4a9a6579fb)

**Host Environment (please complete the following information):**
 -  Arch: amd64
 - OS: Ubuntu Server
 - Version 22.04
 - Docker Version: 3.1.2

PassiveLemon:
Could you report what is in the volume, if anything at all?

LeonardoIz:
It is empty 
![image](https://github.com/user-attachments/assets/7dc6b887-fc7a-41c5-b3be-5015be1ece97)

PassiveLemon:
The volume is created from the VOLUME instruction in the Dockerfile. If a host mount is not supplied at the same location as the volume, then Docker will create the volume. Unfortunately, we can't create a volume at runtime.

I think I could set up a symlink so that when ZeroTier is used, the data is stored in the Zoraxy volume. I'll look into it

PR #392, opened 2024-11-12
Symlink ZeroTier var to Zoraxy config

PassiveLemon:
https://github.com/tobychui/zoraxy/issues/388

Instead of creating a separate volume for ZeroTier itself, we can symlink its var directory to the Zoraxy config and make use of its volume.

This won't cause a loss of data as any users who mount a directory at `/var/lib/zerotier-one` won't have the symlink. If any user wants to simply their volumes, just move any data from the old ZeroTier mount to `zerotier/` in the Zoraxy config directory.


tobychui (approved):
LGTM

PR #406, opened 2024-12-01
Domain Name Server Option

Sickjuicy:
Adds a function to select DNS servers

### DIFF
--- a/src/mod/acme/acme.go
+++ b/src/mod/acme/acme.go
@@ -177,10 +177,16 @@ func (a *ACMEHandler) ObtainCert(domains []string, certificateName string, email
 	certInfo, err := LoadCertInfoJSON(fmt.Sprintf("./conf/certs/%s.json", certificateName))
 	if err == nil {
 		useDNS = certInfo.UseDNS
-		if dnsServers == "" && len(certInfo.DNSServers) > 0 {
+		if dnsServers == "" && certInfo.DNSServers != nil && len(certInfo.DNSServers) > 0 {
 			dnsServers = strings.Join(certInfo.DNSServers, ",")
 		}
 		propagationTimeout = certInfo.PropTimeout
+	}
+
+	// Clean DNS servers
+	dnsNameservers := strings.Split(dnsServers, ",")
+	for i := range dnsNameservers {
+		dnsNameservers[i] = strings.TrimSpace(dnsNameservers[i])
 	}
 
 	// setup how to receive challenge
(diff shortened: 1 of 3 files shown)


src/mod/acme/acme.go
```diff
@@ -292,12 +321,13 @@ func (a *ACMEHandler) ObtainCert(domains []string, certificateName string, email
...
 
 	// Save certificate's ACME info for renew usage
-	certInfo := &CertificateInfoJSON{
+	certInfo = &CertificateInfoJSON{
 		AcmeName:    caName,
 		AcmeUrl:     caUrl,
 		SkipTLS:     skipTLS,
 		UseDNS:      useDNS,
 		PropTimeout: propagationTimeout,
+		DNSServers:  strings.Split(dnsServers, ","),
```
tobychui: I am a bit worry that a space will exists between the `dnsServers` and comma, specially in cases where the user manually write the config file in vim (or other text editors)
e.g. `dns1.example.com, dns2.example.com`
I guess the better way to do so is iterate and clean the dnsServers entry one by one using something like

```
servers := strings.Split(dnsServers, ",")
    for i := range servers {
        servers[i] = strings.TrimSpace(servers[i])
    }
```

src/mod/acme/acme.go
```diff
@@ -164,15 +173,25 @@ func (a *ACMEHandler) ObtainCert(domains []string, certificateName string, email
 		return false, err
 	}
 
+	// Load certificate info from JSON file
+	certInfo, err := LoadCertInfoJSON(fmt.Sprintf("./conf/certs/%s.json", certificateName))
+	if err == nil {
+		useDNS = certInfo.UseDNS
+		if dnsServers == "" && len(certInfo.DNSServers) > 0 {
```
tobychui: As you have added a new field `DNSservers`, when a user using old cert info config file, the `certInfo.DNSServers` will be `nil`. You will also need to check if `certInfo.DNSServers == nil` before this line.

tobychui (changes requested):
Thanks so much for the PR!

I have read it and you really done a great job adding the new DNS server option for ACME. I have raised a few questions and concern in the PR, maybe you can check and see if you can optimize it further before I merge it into main. 

If you are too busy to update it, I can also change your merge target to a new branch and I will continue your work from there.

@yeungalan as you are the primary author of the ACME module, I guess you shd take a look at the changes as well. Apologies to bother you during the thanksgiving holiday, you can PM me directly when you are free.

Sickjuicy:
I updated it and tested it 

normal behavior 
"dnsServers":["ns1.example.com","ns2.example.com"]}



Something put in custom 

dnsServers":[" ns1.example.com  , ns2.example.com"]}

will be updated to after starting the renew process to
"dnsServers":["ns1.example.com","ns2.example.com"]} 

the same if the variable does not yet exist 
result: 

"dnsServers":[""]} 

it will than just uses the standard resolver from google and cloudflare

tobychui (approved):
LGTM, @yeungalan please have a look and let me know if it is ready to merge, thanks!

yeungalan (approved):

tobychui (approved):

PR #421, opened 2024-12-09
Add authelia-verify support

7brend7:
Add authelia integration:
every request proxied by zoraxy is checked for host verification in authelia. If the verification fails, the user is redirected to authelia, where a cookie token is set after login. 
This cookie is domain-specific, so it’s crucial that both authelia and the target application are under the same domain level. Once the cookie is set, the verification process should pass.

https://github.com/user-attachments/assets/d8b644ef-507d-4cc5-a56b-a17a62eb2204

### DIFF
--- a/src/mod/dynamicproxy/authelia.go
+++ b/src/mod/dynamicproxy/authelia.go
@@ -1,0 +1,57 @@
+package dynamicproxy
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+func (h *ProxyHandler) handleAutheliaAuthRouting(w http.ResponseWriter, r *http.Request, pe *ProxyEndpoint) error {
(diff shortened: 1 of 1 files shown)


XHyperDEVX:
this looks great! thanks for this pr!
just out of interest, can this be merged like this? because actually this should be the “SSO/Oauth” tab (or not?)

7brend7:
@XHyperDEVX it can be merged since it doesnt break/change anything.... for my usage - I've replaced BasicAuth to make it work, but for MPV we need UI to enable that and input to provide authelia url.

tobychui (approved):
LGTM, for Authelia users information, this is just a function in a dedicated file and it is not ready to be used. Even if the PR is getting merged into the dev branch, building the dev branch still won't expose this feature until I have finished the integration.

PR #439, opened 2024-12-25
V3.1.5

tobychui:
- Fixed hostname case sensitive bug
- Fixed ACME table too wide css bug
- Fixed HSTS toggle button bug
- Fixed slow GeoIP resolve mode concurrent r/w bug
- Added close connection as default site option
- Added experimental authelia support
- Added custom header support to websocket
- Added levelDB as database implementation (not currently used)
- Added external GeoIP db loading support
- Restructured a lot of modules


tobychui:
Hi @PassiveLemon ,

In this new version, the following two new flags are added

```
-db string
        Database backend to use (leveldb, boltdb, auto) Note that fsdb will be used on unsupported platforms like RISCV (default "auto")

-update_geoip
        Download the latest GeoIP data and exit
```

In which I would recommend running `-update_geoip` on the first startup to keep the geoip db up to date (or maybe add a manual update option as well? I will leave this to your implementation) . 

As I have changed quite a lot of code due to the new introduction of Authentication Provider architecture, maybe you could do a docker stability test as well just to make sure it works reliably? Many thanks!!!

PR #544, opened 2025-02-16
V3.1.8

tobychui:
- Exposed timeout value from dpcore to UI
- Added active load balancing (if uptime monitor is enabled on that rule)
- Refactorized io stats and remove dependencies over wmic
- Removed SMTP input validation
- Fixed sticky session bug
- Fixed passive load balancer bug
- Fixed dockerfile bug


tobychui:
I just notice this is the 100th PR, cool!

PR #545, opened 2025-02-16
Fixed lets encrypt ACME fail bug

tobychui:
### Change log
- Updated lego to fix ACME issue on lets encrypt

@yeungalan I think you should consider re-factorize the ACME module as it started to look like a bunch of expired spaghetti with a few bugs stuck inside. While debugging the module, I found that most of the starting parameter of the module cannot effectively change how the module works (e.g. change the CA link in the main scope from production to staging directory, the ACME module still request from the production server). 

If you need assistant, please let me know so we can discuss the new structure in details.


varmint708:
lego [4.22.2 ](https://github.com/go-acme/lego/releases/tag/v4.22.2)is available which adds support for [myaddr](https://github.com/go-acme/lego/pull/2411) ,so if lego is being updated, can it be updated to latest 4.22.2 instead of 4.21 and support for myaddr be added? Thanks.

tobychui:
> lego [4.22.2 ](https://github.com/go-acme/lego/releases/tag/v4.22.2)is available which adds support for [myaddr](https://github.com/go-acme/lego/pull/2411) ,so if lego is being updated, can it be updated to latest 4.22.2 instead of 4.21 and support for myaddr be added? Thanks. 

You gonna ask @yeungalan for this matter. 

Fyi, after the update we still need to follow up with the code generator which creates the UI entry for the ACME snippet. A new DNS support might take a few weeks at least after this PR is merged.

yeungalan (approved):

PR #593, opened 2025-03-21
Update logger to include UserAgent

Raithmir:
I've never used Go before but this seemed like a simple change so figured I'd create a pull request for it. Tested locally and seems to work as expected.

The inclusion of UserAgent would allow Crowdsec to filter on known bad UserAgent strings.


tobychui (approved):
LGTM

PR #601, opened 2025-03-25
V3.2.0

tobychui:
- Removed GAN and moved to plugin
- Added upnp port forwarder plugin
- Added loopback detection to whitelist
- Completed plugin prototype interface
- Updated dpcore concurrent implementation
- Merged traffic log update
- Updated homepage
- Fixed memory leak on netstat
- Fixed and updated minor dark theme color pallets


tobychui:
Hi @PassiveLemon,

I have removed some of the start flags related to global area network in this revision, so some docker file update might be need here. I guess you might also want to add the ztnc plugin into the plugin folder as well (if you plan to continue be compatible with older version of Zoraxy which supports zerotier out of the box). 

I plan to release the coming versions (v3.2.0 to v3.2.2) as pre-releases, so you can take your time for supporting the new plugin system in docker. Thanks!

PassiveLemon:
As far as I could find, `-ztauth` was the only removed flag, and there were no other additions, is this correct?

How does the plugin system work? Looking in the example dir, I see a script to build all of the plugins, can this be done at runtime or do they need to be built and then packaged? Does the user need to reference this anywhere or does Zoraxy discover them automatically?

Also with the addition of the ztnc plugin, does ZeroTier still need to be installed in the container?

PassiveLemon:
I did some experimenting and got something that works (I think).
I noticed though that the plugin UI's aren't styled and that the ztnc plugin doesn't pick up previous GAN config and the "Create New Network" button doesn't do anything.

The plugins needing to be built before the server starts poses some issues, but I should be able to get a solution worked out.

tobychui:
> With the addition of the ztnc plugin, does ZeroTier still need to be installed in the container?

Yes, it is just the same UI for ZeroTier but moved away from the core.

> I noticed though that the plugin UI's aren't styled and that the ztnc plugin doesn't pick up previous GAN config and the "Create New Network" button doesn't do anything.

Yes, as the previous config is stored in Zoraxy database, there is no way to let a plugin to tab directly into that for security reasons. Though, I might need to take a look at the "button not working" issue later.

> May I request a flag to set the plugins directory? 

Hmm, that should be technically feasible. I will see how can I change the architecture a bit to support user customizable paths for plugin folder.

PassiveLemon:
I've actually come up with a solution that doesn't require the plugin directory flag.

I believe I found a problem in the `build_all.sh` script:
```sh
echo "Copying zoraxy_plugin to all mods"
for dir in ./*; do
    if [ -d "$dir" ]; then
        cp -r ../mod/plugins/zoraxy_plugin "$dir/mod"
    fi
done
...
```
It tries to find `../mod/plugins/zoraxy_plugin` but if I were to cd into the directory with the script, then that relative path doesn't exist. What I suspect is that `zoraxy_plugin` is supposed to be in the plugins directory and then copied to all plugins as some sort of "standard library" for Zoraxy.

Under that assumption, this is my fix:
```sh
#!/bin/bash
# This script builds all the plugins in the current directory

echo "Copying zoraxy_plugin to all mods"
for dir in ./*; do
    if [ -d "$dir" ] && [ ! "$dir" == "./zoraxy_plugin" ]; then
        cp -r ./zoraxy_plugin "$dir/mod"
    fi
done


# Iterate over all directories in the current directory
echo "Running go mod tidy and go build for all directories"
for dir in ./*; do
    if [ -d "$dir" ] && [ ! "$dir" == "./zoraxy_plugin" ]; then
        echo "Processing directory: $dir"
        cd "$dir"
        
        # Execute go mod tidy
        echo "Running go mod tidy in $dir"
        go mod tidy
        
        # Execute go build
        echo "Running go build in $dir"
        go build
        
        # Return to the parent directory
        cd ..
    fi
done

echo "Build process completed for all directories."
```

PassiveLemon:
Oh yeah, in regards to the ztnc plugin button not working, that also applies to the upnp "Search again" and "Add rule". Firefox console returns this for ztnc, but a similar error appears for the others:
```
Uncaught ReferenceError: $ is not defined
    addGANet http://localhost:8005/plugin.ui/org.aroz.zoraxy.ztnc/:117
    onclick http://localhost:8005/plugin.ui/org.aroz.zoraxy.ztnc/:1
org.aroz.zoraxy.ztnc:117:13
```

tobychui:
Hey @PassiveLemon 

Thank you for pointing out the bug in the plugin UI. In fact after a week of digging and tracing, the issue was found originated from a much larger architectural design issue in the embedded web resources file system used in production. I have spent some time to upgrade the webmin UI router design, so now the issue should be fixed (with an added benefit of the web UI is served significant faster than the original implementation). 

The `-plugin` flag is also added for custom plugin folder. I don't really have time to test it this weekend but from my quick test it should be working. You mind help to test and review this PR again?

PassiveLemon:
UI does look correct now, however I noticed that X buttons in the ztnc plugin still don't work, at least the network and networks member X buttons don't, making it impossible to remove networks.

Everything else appears to work though, including the -plugin flag.

tobychui:
@PassiveLemon Cool! 

Yes, I do notice the issue with the ztnc remove button and I have already fixed that in the offical plugin repo over here.
https://github.com/aroz-online/zoraxy-official-plugins

You need editing permission over the new plugin repo?

PassiveLemon:
I don't see anything that I would need access to so you don't need to worry about that for now

PR #623, opened 2025-04-05
Dockerfile: fix path error

pheiduck:


PassiveLemon:
The example directory needs to be copied to the docker/ subdirectory (similar to the src/ subdirectory) because Docker build can't access files outside the directory of the Dockerfile, I'll add a note

PR #628, opened 2025-04-06
Refactor Dockerfile

PassiveLemon:
- Switch back Alpine (I don't get unrecognized symbol errors anymore)
- Download zoraxy_plugin from the plugins repository
- Fix some ZeroTier related issues

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,5 @@
 ## Build Zoraxy
-FROM docker.io/golang:bookworm AS build-zoraxy
+FROM docker.io/golang:alpine AS build-zoraxy
 
 RUN mkdir -p /opt/zoraxy/source/ &&\
     mkdir -p /usr/local/bin/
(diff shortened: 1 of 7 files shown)


docker/Dockerfile
```diff
@@ -15,37 +15,53 @@ RUN go mod tidy &&\
...
+
+RUN mkdir -p /opt/zoraxy/zoraxy_plugin/
+
+RUN apk add --update --no-cache git
+
+WORKDIR /opt/zoraxy/
+
+RUN git clone https://github.com/aroz-online/zoraxy-official-plugins &&\
+    #     v this path with likely need to be updated
+    cp -r ./zoraxy-official-plugins/src/ztnc/mod/zoraxy_plugin/ /opt/zoraxy/zoraxy_plugin/
```
PassiveLemon: Still not sure if `zoraxy-official-plugins/src/ztnc/mod/zoraxy_plugin/` is going to stay so I'll need some information on where the most stable path should be.

tobychui: I am still working on the plugin manger architecture and I am not sure about this too. But I can try to not change this path if possible in the coming releases if that is something you preferred.

PassiveLemon: I think the zoraxy_plugin should have its own place outside of the plugins themselves, that would be ideal. Maybe a directory along-side the other plugins?

tobychui: Not sure what you mean. Can you give an example structure or a diagram of what is your preferred way for storing the system / official plugins that is easy to work with in docker?

PassiveLemon: In the zoraxy-official-plugins repository:
```
src/
    ztnc/
    zoraxy_plugin/
    ... (other plugins)
```
I'm mainly just looking for a dedicated place that is unlikely to suddenly change

tobychui: @PassiveLemon Hmm I don't really get it. The zoraxy-offical-plugin repo only stores Zoraxy's official plugins. The src/ztnc path should be stable and other plugins will be added as the structure you described. So in the future the official plugin repo will look something like this.

```
src/
    ztnc/
    upnp/
    zoramonkey/
    . . . (other plugins)
directories/
    index.json (contains the download paths for the release binary and version info of the plugins)
```

I guess we are suggesting the same structure here?

PassiveLemon: If the ztnc path is expected to be stable, then I'll just keep it where it currently is

Other question: If each plugin has the zoraxy_plugin inside of it, what's the purpose of the "Copying zoraxy_plugin to all mods" part of the build_all.sh script?

tobychui: The zoraxy_plugin is the IPC library of zoraxy plugins where it is used by the plugins to "talks" to the current version of Zoraxy while keeping the API used by the plugin unchanged. Different versions of Zoraxy should came with its own "zoraxy_plugin" library and that is where it needed to be updated in the build script. 

I will try to keep the IPC unchanged (so in the future the zoraxy_plugin folder copy & recompile should be done in a minimum required interval) , but it is really hard to affirm you on this right now as I am still tinkering with the plugin manager design (for Zoraxy) as well as the events pub / sub pipeline (for plugins).

PassiveLemon:
I think this is ready to be merged,

Merge https://github.com/tobychui/zoraxy/pull/643 first and then I'll make any necessary fixes to this one

tobychui:
Ok, #643 has been merged. 

Let me know when this is ready for review and I will merge this into main and sync with the dev branch later during the weekend 👍🏻

PassiveLemon:
Can't find any issues, read for review

tobychui (approved):

PR #635, opened 2025-04-11
Update httprp.html

Lvshujun0918:
When using not standard proxy port, add to the a link to avoid redirect to the wrong path.

### DIFF
--- a/src/web/components/httprp.html
+++ b/src/web/components/httprp.html
@@ -91,102 +91,124 @@
             }else{
                 //Sort by RootOrMatchingDomain field
                 data.sort((a,b) => (a.RootOrMatchingDomain > b.RootOrMatchingDomain) ? 1 : ((b.RootOrMatchingDomain > a.RootOrMatchingDomain) ? -1 : 0))
-                data.forEach(subd => {
-                    let subdData = encodeURIComponent(JSON.stringify(subd));
-                    
-                    //Build the upstream list
-                    let upstreams = "";
-                    if (subd.ActiveOrigins.length == 0){
-                        //Invalid config
-                        upstreams = `<i class="ui yellow exclamation triangle icon"></i> No Active Upstream Origin<br>`;
-                    }else{
-                        subd.ActiveOrigins.forEach(upstream => {
-                            console.log(upstream);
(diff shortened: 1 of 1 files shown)


tobychui:
@Lvshujun0918 Your linter seems changed a lot of lines that is not related to this PR and that is really hard to read. Can you talk about what you have changed and which lines they are located in your PR?

Lvshujun0918:
To get the port info, I need to ajax get /api/status. So I use $.get wrap the previous code.
```javascript
$.get("/api/proxy/status", function(data_1) {
                    if (data_1.error !== undefined) {
                        //Cannot access API, don't change port.
                        proxy_port = `443`;
                    } else {
                        //Use port in the API
                        proxy_port = data_1.Option.Port;
                    }
//the previous code
});
```
So many indents are added, actually just a few new code

Lvshujun0918:
and by the way, this api will return many unused info, which may impact performance. Maybe a api only for getting port should be added.

Lvshujun0918:
Affected Lines
94-104
162-165
172-182
231-238
240-242

tobychui:
Hi @Lvshujun0918 

Ok, I have read your changes. But I don't think the port number should be displayed in the HTTP proxy list, as the port is technically not a part of the hostname (or alias). 

But for the link click behavior, yes, I think those changes are valid fixes (i.e. You should not see the port number in the list as it is not part of a host name, but when a user click on the link, they should be redirected / pop up to the `//host:port` instead of just `//host` for better user experience purpose)

Btw, there is a variable named `currentListeningPort` in status.html which is in the same scope of the `httprp.html`. I think you can use that instead of making an extra GET request to the server to get the current listening port number. Since this implementation is not quite optimized for merging, I will close it and implement a quick patch instead.

Lvshujun0918:
got it. that's fine.

tobychui:
You can take a look at the implementation over [here](https://github.com/tobychui/zoraxy/commit/0c8dfd8aa0fdac6e8592a52a6c9db84cde1bfe3c). This should solved the same issue you are trying to tackle in this PR.

Lvshujun0918:
yeah, it's much more simple.

PR #641, opened 2025-04-21
fix(authelia): original headers

### LINKED ISSUE #502: [ENHANCEMENTS]  Authelia only for Alias
**Is your feature request related to a problem? Please describe.**

Not a problem, just a security issue: I created an access-filter (whitelist) with my internal ips (docker network & internal lan) so i can reach some urls only from internal. 

**Describe the solution you'd like**

it would be nice to be able to create an proxy alias and attach the authelia function ony to this alias  - like service.domain.com is filtered and only reachable from inside my lan (without authelia), but the alias ext.service.domain.com will be reachable from extern, but is first catched from authelia.

**Describe alternatives you've considered**

i guess a simple solution would be to just create a new proxy host accordingly or maybe just the possebility to edit the set_real_ip or real_ip_header
stuff ? 

**Additional context**
yes, i know it "could" be possibel to configure authelia to do the described above, but depending on your infrastructure it ist not an simple task. (nat reflection, firewall, redirects ...)

### LINKED ISSUE #577: [HELP] API bypassing with Autheila, path=/api/verify
I know this is the Zoraxy Github and not the Authelia Github.

**What happened?**
I am running into some issues when attempting to configure API bypassing in the Authelia config.

**Describe what you have tried**
I have tried multiple different regexs to detect API calls.
Bypassing the entire subdomain work. but not when using ` resources: - '^/api(?:[/?].*)?$'` for example. I have tried many different versions of regex for this.

**Describe the networking setup you are using** 
- Are you using the Docker build of Zoraxy? [yes (with docker setup & networking config attach) /no]
    Yes, Docker in Unraid, base config, it is all running on 10.0.1.7.
- Your Zoraxy version? [e.g. 3.0.4]
     v.3.1.9
- Are you using Cloudflare? [yes/no]
     DNS Proxy for external, no tunneling. Using a split DNS for internal traffic.
- Are your systems hosted under a NAT router? [e.g. yes, with subnet is e.g. 192.168.0.0/24 and include port forwarding config if any]
     10.0.1.0/24, 10.0.1.7 pf-> 443.

**Additional context**
I reached out for some help on different platforms and read through extensive documentation, and I was told this: _"Zoraxy is not officially supported by authelia. Authelia supports the following Authz mechanisms ForwardAuth, ExtAuthz, and AuthRequest as well as a legacy implementation. If zoraxy supports any of those it should work with authelia"_ 
I am getting this error from Authelia: 
`time="2025-03-05T20:39:41-05:00" level=info msg="Access to https://sonarr.mydomain.ishere (method POST) is not authorized to user <anonymous>, responding with status code 401" method=POST path=/api/verify remote_ip=xxx.xxx.xxx.xxx` (Using Sonarr here for example)
Seeing `path=/api/verify ` and from the above comment, I am wondering if there is something that I am missing in Zoraxy UI or Zoraxy Config regarding the above mentioned comment. Everything else works as expected.
Attached is my Authelia config as well. Please let me know if you need any further information or testing from me.

[Authelia_Config _redacted.txt](https://github.com/user-attachments/files/19110544/Authelia_Config._redacted.txt)

james-d-elliott:
This fixes the original headers imparted to Authelia, including adding the X-Forwarded-For header, X-Fowarded-Method, and fixing the URI which was absent from the X-Original-URL header. This is also solved by #640 but this change is not as drastic and is not breaking.

Fixes #502, Fixes #577

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,9 +199,9 @@ Loopback web SSH connections, by default, are disabled. This means that if you a
 
 Some section of Zoraxy are contributed by our amazing community and if you have any issues regarding those sections, it would be more efficient if you can tag them directly when creating an issue report.
 
-Authelia Support added by @7brend7
-Authentik Support added by @JokerQyou
-Docker Container List by @eyerrock
+- Authelia Support added by [@7brend7](https://github.com/7brend7)
+- Authentik Support added by [@JokerQyou](https://github.com/JokerQyou)
+- Docker Container List by [@eyerrock](https://github.com/eyerrock)
 
 Thank you so much for your contributions!
 
--- a/README.md
+++ b/README.md
@@ -195,6 +195,16 @@ Loopback web SSH connections, by default, are disabled. This means that if you a
 ./zoraxy -sshlb=true
 ```
 
+## Community Maintained Sections
+
+Some section of Zoraxy are contributed by our amazing community and if you have any issues regarding those sections, it would be more efficient if you can tag them directly when creating an issue report.
+
+Authelia Support added by @7brend7
+Authentik Support added by @JokerQyou
+Docker Container List by @eyerrock
+
+Thank you so much for your contributions!
+
 ## Sponsor This Project
 
 If you like the project and want to support us, please consider a donation. You can use the links below
--- a/src/mod/auth/sso/authelia/authelia.go
+++ b/src/mod/auth/sso/authelia/authelia.go
@@ -3,9 +3,10 @@ package authelia
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
+	"net"
 	"net/http"
 	"net/url"
+	"strings"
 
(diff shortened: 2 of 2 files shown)


tobychui:
Ok, this PR confuse me a bit. From what I see in the changes, I guess this should overlay on top of #640 right?

james-d-elliott:
The intent is a stop-gap for the existing issues depending on how long #640 may take. It fixes the original implementation which was missing the path from the X-Original-URL header, and was also missing the relevant headers to communicate the relevant request method and IP header of the client (X-Forwraded-Method and X-Forwarded-For respectively).

tobychui:
Ok, this make sense. I will rebase this into the next release branch. Thanks for the great work!

tobychui (approved):

PR #644, opened 2025-04-27
V3.2.1

tobychui:
- Merged in authentik forward auth support
- Merged IPv6 whitelist patch
- Added support for basic per host name statistic
- Added experimental plugin store
- Added `$remote_ip` in custom header that filters port number from `$remote_addr`
- Fixed origin is not populated in log bug
- Fixed redirection location rewrite bug


tobychui:
Hey @PassiveLemon, just to let you know that there is a minor update to the start flag that added the `-dev` flag. 
By default the -dev flag is set to false to use the embedded web resources for the webmin panel, so I guess the dockerfile should be fine without supporting it.

PR #645, opened 2025-04-27
feat: forward auth

james-d-elliott:
This adds basic support for forwarded authentication similar to caddy and traefik. This replaces Authelia SSO as it effectively covers exactly the same use cases.


src/mod/update/v315/typedef315.go
```diff
@@ -59,7 +59,7 @@ type AuthProvider int
 const (
 	AuthProviderNone AuthProvider = iota
 	AuthProviderBasicAuth
-	AuthProviderAuthelia
+	AuthProviderForward
```
tobychui, line 62: I think this no need to be changed. Instead, a new v322/typedef321.go should be added to map `AuthProviderAuthelia` to `AuthProviderForward` (which can be added later after merge)

tobychui (changes requested):
Ok just minor issue here, can you restore the change on typedef315.go? I will add another automated upgrader for your new config structure.

james-d-elliott:
I've done this, but I accidentally PR'd the wrong branch. Opening again.

Issue #364, opened 2024-10-29
[ENHANCEMENTS] Add support for PROXY PROTOCOL


xenadmin:
**Is your feature request related to a problem? Please describe.**
HAPROXY, NGINX, Caddy and Traefik all support PROXY PROTOCOL.
Nginx Proxy Manager and Zoraxy don't.
It would be nice, if the two Docker Reverse Proxys with a beginner-friendly GUI, would support it. 

**Describe the solution you'd like**
I would like for Zoraxy to support PROXY PROTOCOL for example for:
https://stalw.art/docs/server/reverse-proxy/proxy-protocol

**Describe alternatives you've considered**
Not using Zoraxy ;-)
Currently, the typical alternative would be using Caddy or Traefik.
Or classic NGINX or HAPROXY. 

**Additional context**
#363

Thaboman1:
+1
This would be great. I'm using Traefik behind Zoraxy.

CalebSnell:
+1
Would like to use this with a MC server that needs proxy protocol for analytics.

tobychui:
Experimental implementation added to v3.2.5 release

PR #718, opened 2025-07-02
Add Proxy Protocol V1 option in TCP Stream Proxy and update Stream Proxy UI

jemmy1794:
### Summary
- Added support for Proxy Protocol V1 option in TCP Stream Proxy.
- Modified Stream Proxy UI to include the new Proxy Protocol toggle.

### Related Issue
References #364


tobychui (approved):
Great work! I guess this module get so much config, it is time to create a struct for it.  Let me merge this first and draft the new config struct for you.

PR #719, opened 2025-07-03
Add EnableLogging to Stream Proxy for log control

jemmy1794:
- Add `EnableLogging` to control TCP/UDP Connection logs to reduce log latency.
- Add `Enable Logging` Option in Stream Proxy rule.
- Update Stream Proxy UI.


tobychui (approved):
LGTM

PR #723, opened 2025-07-06
fix empty sso advanced parameters

7brend7:
Fix fetching of SSO Advanced Options: an empty string results in `[0] = ""`, leading to incorrect behavior when checking `len() === 0`

### DIFF
--- a/src/mod/auth/sso/forward/forward.go
+++ b/src/mod/auth/sso/forward/forward.go
@@ -58,11 +58,20 @@ func NewAuthRouter(options *AuthRouterOptions) *AuthRouter {
 	options.Database.Read(DatabaseTable, DatabaseKeyRequestIncludedCookies, &requestIncludedCookies)
 	options.Database.Read(DatabaseTable, DatabaseKeyRequestExcludedCookies, &requestExcludedCookies)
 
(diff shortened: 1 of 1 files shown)


7brend7:
@james-d-elliott @tobychui had an issue with pihole and zoraxy login (zoraxy fails with CSRF errors)... now everything works even with all  advanced options as empty.

src/mod/auth/sso/forward/forward.go
```diff
@@ -58,11 +58,20 @@ func NewAuthRouter(options *AuthRouterOptions) *AuthRouter {
...
-	options.RequestIncludedCookies = strings.Split(requestIncludedCookies, ",")
-	options.RequestExcludedCookies = strings.Split(requestExcludedCookies, ",")
+	// Helper function to clean empty strings from split results
+	cleanSplit := func(s string) []string {
+		parts := strings.Split(s, ",")
+		if len(parts) == 1 && parts[0] == "" {
+			return []string{}
+		}
+		return parts
+	}
```
james-d-elliott:
```suggestion
	cleanSplit := func(s string) []string {
	        if s == "" {
	          return nil
	        }

		return strings.Split(s, ",")
	}
```

tobychui (approved):
LGTM

PR #724, opened 2025-07-06
sort list of loaded certificates by ExpireDate

7brend7:
Always sort the list of existing certificates by `ExpireDate` in ascending order.


tobychui (approved):
LGTM

PR #728, opened 2025-07-07
Docker changes

PassiveLemon:
No longer build plugins by default:
As there is currently a plugin store in the works, I decided to remove the plugin build script as it largely will not be needed. Users can still use custom plugins, they just will have to be built outside the container.

Switch to a python entrypoint:
It's much easier to work with signals and scripting in python than bash and this can make hardening easier in the future by removing dependency on the shell (Other dependencies need the shell so it cannot be removed yet).

As Zoraxy is the front line of defense for self-hosted applications, security is important. In the future, I would like to completely remove the shell from the Docker image, run Zoraxy as rootless, and potentially use a distroless image. These should offer some substantial security gains, though they will be difficult to accomplish


tobychui:
> this can make hardening easier in the future by removing dependency on the shell

Zoraxy itself uses shell command in its code (and ssh, if WebSSH feature is supported on that platform), so I don't think completely removing shell dependency is possible since Zoraxy itself needs to call some of the services via shell.

The rest of the changes looks good to me. Is this tested and working on your environment? If yes I will be merging this directly.

tobychui (approved):

PassiveLemon:
Yep, I've built and tested to see if I could find any bugs. I don't have an environment to test everything but I didn't notice any problems.

AnthonyMichaelTDM:
> Zoraxy itself uses shell command in its code (and ssh, if WebSSH feature is supported on that platform), so I don't think completely removing shell dependency is possible since Zoraxy itself needs to call some of the services via shell.

Would this still be a problem if the WebSSH feature was reimplemented as a plugin?

PR #739, opened 2025-07-13
fix: typo in dynamic_router.go

AnthonyMichaelTDM:
SniffResultAccpet should be SniffResultAccept


AnthonyMichaelTDM:
It should be fine (since I assume the enums are exported as their constant values (0 and 1, in this case), and the meaning behind each value is unchanged), but that's understandable

AnthonyMichaelTDM:
I'll fork off the dev branch next time

AnthonyMichaelTDM:
I'm going to rebase this on top of the v3.2.5 branch then force push

AnthonyMichaelTDM:
there we go, all better

PR #746, opened 2025-07-18
feat(plugins): Implement plugin API key management and authentication middleware

AnthonyMichaelTDM:
## Motivation

Currently, Zoraxy renders plugins in a webframe and allows various ways of deciding which kinds of requests should be handled by a plugin.

A big limitation of the plugin system is that while plugins can make requests to their own backend, they cannot interract with the zoraxy api in any meaningful capacity.

I'm implementing a crowdsec bouncer for zoraxy, ideally I'd be able to adjust the zoraxy blacklist directly but due to the previously mentioned limitations I had to implement the bouncer as a dynamic capture plugin which is less then ideal for a variety of reasons.

## Purpose

The purpose of this PR is to give plugins scoped access to internal APIs over https.

## Overview

Plugins can declare a list of `PermittedAPIEndpoint`'s in their `IntroSpect`, each must include the endpoint, the http method (e.g. GET, PUSH), and the reason.
Plugins that declare such a list will receive an API key as part of their `ConfigureSpec`

These API keys are generated for each applicable plugin when they start, and are invalidated when the plugin stops.

When making requests to the Zoraxy API, the api key should be included as `Bearer <API_KEY>` in the `Authorization` header.

The `authRouter` has been modified so that handlers for `pluginAccessible` endpoints are wrapped in an additional layer of middleware that will look for the Auth Bearer header, and if its present will validate that the provided API key has access to the endpoint. If the key has access, the normal auth mechanism can be bypassed. If the key is invalid or does not have access, the request fails with a 403 Unauthorized message.

Requests to non-`pluginAccessible` endpoints redirect to the login page as they did before.
Requests to `pluginAccessible` endpoints that are missing the Auth Bearer header are treated as they were before.

## Status

- [x] feature implemented
- [x] example plugin to demonstrate using the feature
- [x] feature integrated into webUI

### Where PermittedAPIEndpoints are displayed in the web UI

image

### UI of the example plugin

image
image

This UI demonstrates that:
- valid requests to permitted endpoints succeed
- invalid (wrong key) requests to permitted endpoints are unauthorized
- valid requests to unpermitted endpoints are unauthorized
- requests to endpoints that aren't plugin accessible fail (you get the login page (normal auth behavior))

## Future work

- [ ] more carefully examine which API are exposed, current implementation is fairly conservative about which endpoints are accessible but may very well be too restrictive
- [ ] add methods to make it easier for plugin authors to make calls to the endpoint (specifically, to automate setting the Auth Bearer header)
- [ ] use a more "industry-standard" method of generating the tokens, currently the token is generated by randomly generating 32 bytes, then hashing that with sha256
- [ ] maybe don't pass the api key in plain text over HTTP? probably doesn't matter since it's all happening over the loopback interface (localhost) but like .. probably still bad practice

### DIFF
--- a/example/plugins/api-call-example/ui.go
+++ b/example/plugins/api-call-example/ui.go
@@ -13,7 +13,7 @@ import (
 func allowedEndpoint(cfg *plugin.ConfigureSpec) (string, error) {
 	// Make an API call to the permitted endpoint
 	client := &http.Client{}
-	apiURL := fmt.Sprintf("http://localhost:%d/api/access/list", cfg.ZoraxyPort)
+	apiURL := fmt.Sprintf("http://localhost:%d/plugin/api/access/list", cfg.ZoraxyPort)
 	req, err := http.NewRequest(http.MethodGet, apiURL, nil)
 	if err != nil {
 		return "", fmt.Errorf("error creating request: %v", err)
@@ -45,7 +45,7 @@ func allowedEndpoint(cfg *plugin.ConfigureSpec) (string, error) {
 func allowedEndpointInvalidKey(cfg *plugin.ConfigureSpec) (string, error) {
 	// Make an API call to the permitted endpoint with an invalid key
 	client := &http.Client{}
-	apiURL := fmt.Sprintf("http://localhost:%d/api/access/list", cfg.ZoraxyPort)
+	apiURL := fmt.Sprintf("http://localhost:%d/plugin/api/access/list", cfg.ZoraxyPort)
 	req, err := http.NewRequest(http.MethodGet, apiURL, nil)
 	if err != nil {
 		return "", fmt.Errorf("error creating request: %v", err)
@@ -98,7 +98,7 @@ func unaccessibleEndpoint(cfg *plugin.ConfigureSpec) (string, error) {
 func unpermittedEndpoint(cfg *plugin.ConfigureSpec) (string, error) {
 	// Make an API call to an endpoint that is plugin-accessible but is not permitted
 	client := &http.Client{}
-	apiURL := fmt.Sprintf("http://localhost:%d/api/proxy/list", cfg.ZoraxyPort)
+	apiURL := fmt.Sprintf("http://localhost:%d/plugin/api/proxy/list", cfg.ZoraxyPort)
 	req, err := http.NewRequest(http.MethodGet, apiURL, nil)
 	if err != nil {
 		return "", fmt.Errorf("error creating request: %v", err)
@@ -232,35 +232,36 @@ func RenderUI(config *plugin.ConfigureSpec, w http.ResponseWriter, r *http.Reque
 		
 		<div class="response-block success">
 			<h3>✅ Allowed Endpoint (Valid API Key)</h3>
-			<p>Making a GET request to <code>/api/access/list</code> with a valid API key:</p>
+			<p>Making a GET request to <code>/plugin/api/access/list</code> with a valid API key:</p>
 			<div class="response-content">
 				` + RenderedAccessListHTML + `
 			</div>
 		</div>
 
-		<div class="response-block error">
-			<h3>❌ Invalid API Key</h3>
-			<p>Making a GET request to <code>/api/access/list</code> with an invalid API key:</p>
+		<div class="response-block warning">
+			<h3>⚠️ Invalid API Key</h3>
+			<p>Making a GET request to <code>/plugin/api/access/list</code> with an invalid API key:</p>
 			<div class="response-content">
 				` + RenderedInvalidKeyResponseHTML + `
 			</div>
 		</div>
 
 		<div class="response-block warning">
-			<h3>⚠️ Disallowed Endpoint</h3>
+			<h3>⚠️ Unpermitted Endpoint</h3>
+			<p>Making a GET request to <code>/plugin/api/proxy/list</code> (not a permitted endpoint):</p>
+			<div class="response-content">
+				` + RenderedUnpermittedResponseHTML + `
+			</div>
+		</div>
+
+		<div class="response-block error">
+			<h3>❌ Disallowed Endpoint</h3>
 			<p>Making a GET request to <code>/api/acme/listExpiredDomains</code> (not a plugin-accessible endpoint):</p>
 			<div class="response-content">
 				` + RenderedUnaccessibleResponseHTML + `
 			</div>
 		</div>
 
-		<div class="response-block warning">
-			<h3>⚠️ Unpermitted Endpoint</h3>
-			<p>Making a GET request to <code>/api/proxy/list</code> (plugin-accessible but not permitted):</p>
-			<div class="response-content">
-				` + RenderedUnpermittedResponseHTML + `
-			</div>
-		</div>
 	</body>
 	</html>`
 	w.Write([]byte(html))
--- a/example/plugins/api-call-example/ui.go
+++ b/example/plugins/api-call-example/ui.go
@@ -1,0 +1,223 @@
+package main
+
+import (
+	"fmt"
+	"html"
+	"net/http"
+	"net/http/httputil"
+	"strconv"
+
+	plugin "aroz.org/zoraxy/api-call-example/mod/zoraxy_plugin"
+)
+
+func allowedEndpoint(cfg *plugin.ConfigureSpec) (string, error) {
+	// Make an API call to the permitted endpoint
+	client := &http.Client{}
+	apiURL := fmt.Sprintf("http://localhost:%d/api/access/list", cfg.ZoraxyPort)
+	req, err := http.NewRequest(http.MethodGet, apiURL, nil)
+	if err != nil {
+		return "", fmt.Errorf("error creating request: %v", err)
+	}
+	// Make sure to set the Authorization header
+	req.Header.Set("Authorization", "Bearer "+cfg.APIKey) // Use the API key from the runtime config
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("error making API call: %v", err)
+	}
+	defer resp.Body.Close()
+
+	// Check if the response status is OK
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("received non-OK response status %d", resp.StatusCode)
+	}
+
+	respDump, err := httputil.DumpResponse(resp, true)
+	if err != nil {
+
+		return "", fmt.Errorf("error dumping response: %v", err)
+	}
+
+	return string(respDump), nil
+}
+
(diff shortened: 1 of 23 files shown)


AnthonyMichaelTDM:
It's a pretty big PR, sorry about that, but about 2/3rds of the additions (1100 lines) are from the example plugin so it looks worse than it is

AnthonyMichaelTDM:
regarding

> @tobychui https://github.com/tobychui/zoraxy/discussions/738#discussioncomment-13801653
> In fact I kinda want to work on a restful API sub system for Zoraxy (for automation and testing), maybe these two features can share the same token db?

That should be possible, currently a given API key lives only as long as the plugin is running, no attempt is made at persisting the token DB since tokens are regenerated at runtime.
What this does mean is that a restful API sub-system could be very nicely implemented *as a plugin* with the only modifications to the source-code being to make more endpoints plugin-accessible.

Note: the distinction between `pluginAccessible` and non `pluginAccessible` endpoints is motivated by the assumption that some endpoints should only ever be accessible to the actual admin (e.g. changing passwords, (un)installing plugins (this one especially since if there was an exploit in the plugin-auth system, being able to install and run plugins could be a (wormable?) RCE vulnerability), etc.)

example/plugins/api-call-example/ui.go
AnthonyMichaelTDM, line 1: this plugin is basically a test that the plugin-auth works under a few specific scenarios.

This plugin is only permitted to access the `/api/access/list` endpoint

src/api.go
AnthonyMichaelTDM, line 1: this is where endpoints are marked as either `pluginAccessible` or not

src/mod/auth/auth.go
AnthonyMichaelTDM, line 1: the plugin auth middleware is added to the existing auth agent

src/mod/auth/plugin_apikey_manager.go
```diff
@@ -0,0 +1,152 @@
...
+		APIKey:             apiKey,
+		PermittedEndpoints: permittedEndpoints,
+		CreatedAt:          time.Now(),
+	}
+
+	// Store the API key
+	m.keys[apiKey] = pluginAPIKey
+
+	return pluginAPIKey, nil
+}
```
AnthonyMichaelTDM, line 63: here's how keys are generated

src/mod/auth/plugin_apikey_manager.go
```diff
@@ -0,0 +1,152 @@
...
+
+	// Check if the endpoint is permitted
+	for _, permittedEndpoint := range pluginAPIKey.PermittedEndpoints {
+		if permittedEndpoint.Endpoint == endpoint && permittedEndpoint.Method == method {
+			return pluginAPIKey, nil
+		}
+	}
+
+	return nil, fmt.Errorf("endpoint not permitted for this API key")
+}
```
AnthonyMichaelTDM, line 93: and how access is validated

src/mod/auth/plugin_middleware.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		// Add plugin information to the request context
+		r.Header.Set("X-Zoraxy-Plugin-ID", pluginAPIKey.PluginID)
+		r.Header.Set("X-Zoraxy-Plugin-Auth", "true")
+
+		// Call the original handler
+		handler(w, r)
+	}
+}
```
AnthonyMichaelTDM: this is where api requests coming from plugins are validated, the `X-Zoraxy-Plugin-ID` header is used to communicate the middleware's decision to the handler (which is the previous auth wrapper)

src/mod/auth/router.go
AnthonyMichaelTDM, line 1: this is basically part 2 of the auth flow.
if the header is present, and valid, bypass the auth agent.

src/mod/auth/router.go
```diff
@@ -34,31 +34,28 @@ func (router *RouterDef) HandleFunc(endpoint string, handler func(http.ResponseW
...
-				authAgent.HandleCheckAuth(w, r, func(w http.ResponseWriter, r *http.Request) {
-					handler(w, r)
-				})
-			} else {
-				handler(w, r)
-			}
+	// if the endpoint is supposed to be plugin accessible, wrap it with plugin authentication middleware
+	if pluginAccessible {
+		authWrapper = router.option.AuthAgent.PluginAuthMiddleware.WrapHandler(endpoint, authWrapper)
+	}
```
AnthonyMichaelTDM: if this is a plugin endpoint, wrap it in the plugin auth middleware

src/mod/plugins/lifecycle.go
AnthonyMichaelTDM, line 1: An API key is generated when the plugin starts, and revoked when it exits

src/mod/auth/router.go
AnthonyMichaelTDM: maybe worth also validating that the requestAddr is coming from localhost?

src/mod/plugins/zoraxy_plugin/zoraxy_plugin.go
AnthonyMichaelTDM, line 1: new spec fields

AnthonyMichaelTDM (commented):
This "review" is just a code tour, I do this with big PRs to make them easier to digest

AnthonyMichaelTDM:
> regarding
> 
> > @tobychui [#738 (reply in thread)](https://github.com/tobychui/zoraxy/discussions/738#discussioncomment-13801653)
> > In fact I kinda want to work on a restful API sub system for Zoraxy (for automation and testing), maybe these two features can share the same token db?
> 
> That should be possible, currently a given API key lives only as long as the plugin is running, no attempt is made at persisting the token DB since tokens are regenerated at runtime. What this does mean is that a restful API sub-system could be very nicely implemented _as a plugin_ with the only modifications to the source-code being to make more endpoints plugin-accessible.
> 
> Note: the distinction between `pluginAccessible` and non `pluginAccessible` endpoints is motivated by the assumption that some endpoints should only ever be accessible to the actual admin (e.g. changing passwords, (un)installing plugins (this one especially since if there was an exploit in the plugin-auth system, being able to install and run plugins could be a (wormable?) RCE vulnerability), etc.)

Actually, expanding on this further, I wonder how viable it would be to have the example plugin read a list of endpoints to register and endpoints to hit from a config file and then dynamically create that "report" in the UI based on that list

Thinking about it more, yeah, I could definitely do that. The real question then is whether to include that in this PR or wait for this to be merged and do it in a new PR

The format could be something like:

```yml
permitted-endpoints:
    - endpoint: /api/access/list
      method: GET
      reason: used as an example of a permitted and pluginAccessible endpoint
    - endpoint: /api/cert/tls
      method: GET
      reason: used as an example of a "permitted" but not pluginAccessible 

endpoints-to-hit:
    - endpoint: /api/access/list
      expected-result: success
    - endpoint: /api/access/list
      api-key-override: invalid-key
      expected-result: redirect
    - endpoint: /api/cert/tls
      expected-result: redirect
    - endpoint: /api/access/list
      method-override: HEAD
      expected-result: unauthorized
    - endpoint: /api/acme/listExpiredDomains
      expected-result: unauthorized
    - endpoint: /api/proxy/list
      expected-result: unauthorized
```

which could allow the plugin to be used as a kind of test-suite for the rest API

The webUI of the plugin would generate those blocks automatically based on the response (green for success, red for redirect, yellow for unauthorized, purple for other) and report whether the response matched the expectation

src/mod/auth/router.go
tobychui: I guess a whitelist for allowed IPs will be a better implementation here. Otherwise, someone will eventually create a plugin that accept external connections and convert it to Zoraxy local API calls, which might makes it even harder to maintain.

src/mod/auth/router.go
```diff
@@ -25,7 +25,7 @@ func NewManagedHTTPRouter(option RouterOption) *RouterDef {
 	}
 }
 
-func (router *RouterDef) HandleFunc(endpoint string, handler func(http.ResponseWriter, *http.Request)) error {
+func (router *RouterDef) HandleFunc(endpoint string, handler func(http.ResponseWriter, *http.Request), pluginAccessible bool) error {
```
tobychui: Adding this will break the compatibility with the native `net.http` `HandleFunc` function. I guess it would be better to implement it as alternative API instead (maybe adding a new handle function instead?)

src/api.go
tobychui: This doesn't seems to be a good implementation. Per my comment in auth/router.go, perhaps you can add a new handler in auth module for supporting plugin api route?  We can have another go script (e.g. plugin_api.go) just for binding API functions for plugins.

AnthonyMichaelTDM:
Sure, those comments make a lot of sense.
What path should I put the new endpoints at? I'm thinking something like `/plugin-api/...` or `/plugin/api/...` or maybe `/rest/...`

Also, on that point, which endpoints should be included there? Should I just start with the ones I have marked as pluginAccessible and leave the rest as future work?

I'll probably get started on this tomorrow

src/mod/auth/router.go
AnthonyMichaelTDM: oh I didn't even think about that, great point

AnthonyMichaelTDM:
> I'll probably get started on this tomorrow

nvm, it was easier than I thought, commits incoming

src/mod/auth/router.go
AnthonyMichaelTDM: I'm resolving this since the code being referenced no longer exists

AnthonyMichaelTDM:
@tobychui 
I've implemented your requested changes if you'd like to take a look
(still need to update the example plugin though)

tobychui:
@AnthonyMichaelTDM Cool! Let me release v3.2.5 first and I will fork a v3.2.6 branch and merge this into the new branch. This should give us more time testing it before it got released. Will wait for your example update, let me know when you are ready :)

AnthonyMichaelTDM:
~~ready when you are~~
~~scratch that, somethings wrong. gimme a second~~
okay, fixed. actually ready now

AnthonyMichaelTDM:
Well that was a goofy mistake, all good now
image
image

AnthonyMichaelTDM:
> LGTM, I few minor things that I might wanna change but I will merge this first.

Sounds good, do you want to open an issue for those minor things?

tobychui:
> Sounds good, do you want to open an issue for those minor things?

No, I think I will directly push it into the v3.2.6 branch.

AnthonyMichaelTDM:
Fair enough, in that case I'm going to sign off for the night.

Hope to work on either: resolve the issues causing the CI script I'm adding in #752 to fail (pending a decision on whether you'd prefer those changes done in that PR or in a new one), or creating the example plugin for #753 tomorrow.

tobychui:
@AnthonyMichaelTDM Just wanna ask about the embedded web router HandleFunc issue, I tried to automate the plugin system rebuild using the build all bash script and it seems the restful example in this PR have some issue building.

圖片

I guess the HandleFunc is somehow "gone" during merge or you might have forgotten to commit something?

PR #752, opened 2025-07-20
feat: add a basic CI script

AnthonyMichaelTDM:
This just makes sure the code compiles and passes some static analysis, doesn't run the tests since a lot of them hang on my system.

Worth noting that the static analysis does currently fail, but that can be fixed.

If you want, I can do the fixes in this PR, or put it off for later, I don't mind either way.


AnthonyMichaelTDM:
I'll squash this

AnthonyMichaelTDM:
the dpcore_test.go tests are failing for the following reasons:
- `TestHTTP1p1KeepAlive`: no http server is running
- `TestReplaceLocationHost`: the result doesn't match the expected result, after passing the result through `url.QueryUnescape` they are equal, this explains the confusing error message

the dbleveldb_test.go tests are failing for the following reasons:
- `TestWriteAndRead`: the write isn't synced before attempting to be read (syncing is done on a set interval in a separate go-routine), and it seems there's no mechanism for reading unsynced data.
- `TestListTable`: a little more complex, seems to have multiple errors/bugs piled on top of eachother

The `netutils_test.go` tests seem to both be failing for similar reasons, basically it's failing to create a icmp connection to the IPs that the domains are pointing to:
image

two `acme_test.go` tests are also failing, one seemingly due to a missing file and the other presumably due to malformed input:
image





After we get those tests to pass, staticcheck will fail primarily due to a number of unused functions.

AnthonyMichaelTDM:
Here's a patch that will run an http server for the `TestHTTP1p1KeepAlive` test, I'm not committing it though since I'm unsure what's supposed to be getting tested here

```patch
diff --git a/src/mod/dynamicproxy/dpcore/dpcore_test.go b/src/mod/dynamicproxy/dpcore/dpcore_test.go
index 4e33ffe..b4e18c1 100644
--- a/src/mod/dynamicproxy/dpcore/dpcore_test.go
+++ b/src/mod/dynamicproxy/dpcore/dpcore_test.go
@@ -90,13 +90,33 @@ func TestReplaceLocationHostRelative(t *testing.T) {
 
 // Not sure why this test is not working, but at least this make the QA guy happy
 func TestHTTP1p1KeepAlive(t *testing.T) {
+	// Start a simple HTTP server
+	// This server should have Keep-Alive enabled by default
+	errChan := make(chan error, 1)
+	go func() {
+		http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusOK)
+			w.Write([]byte("Hello, World!"))
+		})
+		err := http.ListenAndServe(":8080", nil)
+		if err != nil {
+			errChan <- err
+		}
+	}()
+	// Give the server a moment to start
+	select {
+	case err := <-errChan:
+		t.Fatal("ListenAndServe: ", err)
+	case <-time.After(2 * time.Second):
+	}
+
 	client := &http.Client{
 		Transport: &http.Transport{
 			DisableKeepAlives: false,
 		},
 	}
 
-	req, err := http.NewRequest("GET", "http://localhost:80", nil)
+	req, err := http.NewRequest("GET", "http://localhost:8080", nil)
 	if err != nil {
 		t.Fatalf("Failed to create request: %v", err)
 	}
@@ -116,7 +136,7 @@ func TestHTTP1p1KeepAlive(t *testing.T) {
 	t.Logf("First request status code: %v", resp.StatusCode)
 	time.Sleep(20 * time.Second)
 
-	req2, err := http.NewRequest("GET", "http://localhost:80", nil)
+	req2, err := http.NewRequest("GET", "http://localhost:8080", nil)
 	if err != nil {
 		t.Fatalf("Failed to create second request: %v", err)
 	}
```

tobychui:
I guess this need to be left for later, since I have my own jenkin server for building (instead of using Github action for the binary build). I will see if I got time to integrate these. Let me change this merge target to main.

AnthonyMichaelTDM:
Is there a way to link jenkins w/ GH so that there's more observability about which tests are failing and such?

tobychui:
Moving this to main so it won't get closed when I remove the v3.2.6 dev branch

Issue #660, opened 2025-05-09
[ENHANCEMENTS] Metrics exposed in Prometheus format


Zydepoint:
Would be great if there is metrics exposed to a /metrics endpoint so that i can scrape it with prometheus and view it with grafana. 

metrics like these for example
- amount of proxies
- targets and uptime/SLA
- tcp redirects
- uptime/health
- network traffic to/from

Issue #461, opened 2025-01-03
[ENHANCEMENTS] Expose Monitoring Data for Prometheus


Trackhe:
It would be nice to get Metrics over an endpoint for Prometheus or similar systems. Instead of only grab them from the logs as i read in other discussions / issues here.

I think that would not have priority because we have some metrics on the Dashboard but it would be nice i you consider it.

PR #753, opened 2025-07-20
feat(plugins): Implement event system w/ POC events

AnthonyMichaelTDM:
## Description

Implements the partially created event system with 3 events implemented as proof of concepts.

The 3 events are:
- `blacklistedIpBlocked`: emitted when a request from a blacklisted IP
- `accessRuleCreated`: emitted when a new access rule is created
- `blacklistToggled`: emitted when the blacklist is toggled for a given access rule

Why these events specifically? Because these are the ones I foresee myself needing in the next version of the zoraxy_crowdsec_bouncer

Events are dispatched via a global event manager `plugins.EventSystem.Emit`

Currently, not tested (the example plugin is the test), ~~partially blocked on #746 since they will both want to make changes to `src/web/snippet/pluginInfo.html`~~

## Status

- [x] feature implemented
- [x] example plugin to demonstrate using the feature
- [x] feature integrated into webUI (new IntroSpec and ConfigSpec fields are displayed to the user)

## Future work

- [ ] implement a lot more events
- [ ] maybe integrate with the `statistic` module?
- [ ] migrate to using UDP for events instead of TCP, this should enable sending events much more frequently (important if we're going to use it for metrics exporters, i imagine)

A sufficiently advanced event system could allow a prometheus metrics exporter to be implemented as a plugin, like was requested in #660 and #461


AnthonyMichaelTDM:
## Example Plugin

The example plugin is subscribed to all 3 currently implemented events and will simply store the event in a list, the WebUI then displays the events in that list. 
In this way, the example plugin is essentially an event log

## Pictures

### When no events have been sent

image

### After some blacklist events have been sent

image

AnthonyMichaelTDM:
## WebUI integration

### when no events are subscribed to

image

### when there are events subscribed to

image

AnthonyMichaelTDM:
@tobychui example plugin and UI integration are complete.

AnthonyMichaelTDM:
meaning the tests I just pushed failed, I'll figure it out

src/mod/plugins/zoraxy_plugin/event.go
AnthonyMichaelTDM, line 1: this probably isn't the *best* way to do things, but idk maybe it is?

AnthonyMichaelTDM:
going to update this branch too

AnthonyMichaelTDM:
I think this is pretty much done as far as the initial implementation goes, is there something you want me to change @tobychui ?

tobychui:
@AnthonyMichaelTDM I am taking a break from coding till 10 Aug. Will follow up this later on. As always, thanks for the PR :)

AnthonyMichaelTDM:
Understood, thanks for letting me know

AnthonyMichaelTDM:
I rebased the branch ontop of v3.2.6

src/accesslist.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/microcosm-cc/bluemonday"
 
 	"imuslab.com/zoraxy/mod/access"
+	"imuslab.com/zoraxy/mod/plugins"
+	"imuslab.com/zoraxy/mod/plugins/zoraxy_plugin"
```
tobychui: Including plugin module here will become hard to maintain. I guess the best way to do so is adding a separate event management modules that consume and emit events for other modules to handle, including the plugin module.

src/mod/access/access.go
```diff
@@ -60,6 +60,7 @@ func NewAccessController(options *Options) (*Controller, error) {
 		//Create one
 		js, _ := json.MarshalIndent(defaultAccessRule, "", " ")
 		os.WriteFile(defaultRuleSettingFile, js, 0775)
+
```
tobychui, line 63: I guess there are some issue with your linter? Please try to not commit changes on file that only add new spaces

src/mod/dynamicproxy/access.go
```diff
@@ -7,6 +7,8 @@ import (
 
 	"imuslab.com/zoraxy/mod/access"
 	"imuslab.com/zoraxy/mod/netutils"
+	"imuslab.com/zoraxy/mod/plugins"
+	"imuslab.com/zoraxy/mod/plugins/zoraxy_plugin"
```
tobychui: The same architectural design issue here.

src/mod/plugins/event_system.go
```diff
@@ -0,0 +1,192 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"imuslab.com/zoraxy/mod/info/logger"
+	zoraxyPlugin "imuslab.com/zoraxy/mod/plugins/zoraxy_plugin"
+)
+
+// eventManager manages event subscriptions and dispatching
+type eventManager struct {
```
tobychui: IMO I think this should be a separate module on itself. This section should only contains the code that recv events from a central event manager and forward it to plugins that listen them.

src/mod/plugins/zoraxy_plugin/event.go
tobychui: You are right, this should be a dedicated module (see my other comments in the code review). 
The zoraxy_plugin should register all events to the central event module and forward them to plugins. 

This seems to be quite some large changes so I guess it would be better for us to draft the events and the module interaction in some sort of design document (like the one I drafted previously for the plugin system) before we start working on this together.

src/start.go
```diff
@@ -347,6 +347,11 @@ func startupSequence() {
 		HotReloadInterval: 5,                  //seconds
 	})
 
+	/*
+		Event Manager
+	*/
+	plugins.InitEventManager(pluginManager.GetPluginByID, SystemWideLogger)
```
tobychui: Maybe this need to be changed to `eventManager.Init()` and later, do the `plugins.SetEventManager(eventManger) ` kind of design

src/mod/plugins/event_system.go
AnthonyMichaelTDM: that makes sense

src/mod/plugins/zoraxy_plugin/event.go
AnthonyMichaelTDM: The thing is that plugins need access to the types, which is most of what this file consists of

AnthonyMichaelTDM: I definitely agree regarding the implementation of the actual event publishing and such (`event_system.go`) though

src/mod/access/access.go
AnthonyMichaelTDM: my bad

src/accesslist.go
AnthonyMichaelTDM: Maybe `imuslab.com/zoraxy/mod/event_system` and `imuslab.com/zoraxy/mod/plugins/zoraxy_plugin/events`?

The reason for keeping specifically `event.go` with the plugins module is that plugins need the types

src/start.go
AnthonyMichaelTDM: alrighty

AnthonyMichaelTDM:
I've moved the event system itself to its own module, but have put off moving `events.go` until we've discussed the implementation more

tobychui:
Ok cool! I will see if I get any time to review and merge this in the coming weekend.

src/accesslist.go
tobychui: That is a good point. But I don't want to mix the scope of plugin and event system. Maybe I can workout something in the CI flow to copy the types from event system to plugin module when build?

src/mod/plugins/zoraxy_plugin/event.go
```diff
@@ -0,0 +1,120 @@
...
+	"fmt"
+)
+
+// EventName represents the type of event
+type EventName string
+
+// EventPayload interface for all event payloads
+type EventPayload interface {
+	// GetName returns the event type
+	GetName() EventName
```
tobychui: I might wanna add a few more methods in the interface, but I guess we can merge this first

tobychui (approved):
Cool, this new architecture looks great. I think I can merge this first and we can work out together in the v3.2.6 branch. What do you think?

src/mod/plugins/zoraxy_plugin/event.go
AnthonyMichaelTDM: what kinds of methods are you thinking of?

src/accesslist.go
AnthonyMichaelTDM: That sounds a bit over-complicated personally, I'm totally down with making it a submodule of zoraxy_plugin (so that code that needs to send events isn't pulling in the entire plugins module to do it), but I think it should be inside zoraxy_plugin

src/mod/plugins/zoraxy_plugin/event.go
tobychui: Like `GetEvenSource()` to see who emitted the payload and `GetPayloadType()` for dumping the original data structure? These seems helpful for making IPC calls.

AnthonyMichaelTDM:
Moved the event types into a submodule of zoraxy_plugin

src/mod/plugins/zoraxy_plugin/event.go
AnthonyMichaelTDM: Hmm, how about we put that off for now. I think we can integrate it later when a use-case pops up (much like "other event types")

tobychui: @AnthonyMichaelTDM I would love to, but it is an interface not a struct. So if a plugin is developed for version `x` and we changed the interface for a bit, a ton of plugins will fail to build on version `x+1` . 

It would be better for us to figure out all the interfaces and try to minimize the change to this from a forward compatibility stand point. Or alternatively, I can label the v3.2.6 as a pre-release and try to figure out the rest of the interfaces before the next release. Which option would you prefer?

AnthonyMichaelTDM: That's a good point, how about we leave it be for now and implement the new functions in a new PR, which we finish before releasing 3.2.6?

src/accesslist.go
AnthonyMichaelTDM: I did move them to a submodule, is that more what you're looking for?

tobychui: Looks good, we will stick with the submodule design 👍

src/mod/plugins/zoraxy_plugin/event.go
AnthonyMichaelTDM: I'll take the emoji as approval, merge, and open an issue for adding the new functions

PR #788, opened 2025-08-23
feat(sso): forward auth body and alternate headers

james-d-elliott:
This implements a minor modification to the forward authz sso where the body can be copied to the auth server and the X-Original-* implementations can be used.

### DIFF
--- a/src/mod/auth/sso/forward/const.go
+++ b/src/mod/auth/sso/forward/const.go
@@ -13,14 +13,21 @@ const (
 	DatabaseKeyRequestHeaders         = "requestHeaders"
 	DatabaseKeyRequestIncludedCookies = "requestIncludedCookies"
 	DatabaseKeyRequestExcludedCookies = "requestExcludedCookies"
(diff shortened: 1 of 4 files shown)


fuad00:
keep it going! really waiting for this one!

tobychui (approved):

obkgroove21:
Hello @james-d-elliott ,

is this a solution for something like this working in Zoraxy?
https://joeeey.com/blog/selfhosting-sso-with-nginx-oauth2-proxy-part-2/

Greetings!

james-d-elliott:
You could use a similar process with Zoraxy I believe. I implemented this based on how Traefik and Caddy do it, and both of those work with OAuth 2.0 Proxy. There is also a toggle that roughly mimics the nginx k8s ingress, so it's probably directly compatible.

obkgroove21:
Hello @james-d-elliott ,

I´m struggeling on how to achieve/transfer the part with the keycloak groups. I already have oauth2proxy running but I don´t know where to put the group for each application.

Do you have any idea and could help me?

PR #804, opened 2025-09-06
V3.2.6

tobychui:
- Optimized SSO / forward auth feature 
- Added plugins API call and example
- Updated lego for DNS challenge
- Added experimental log rotate (default off)
- Updated log viewer UI

### DIFF
--- a/src/mod/info/logger/logger.go
+++ b/src/mod/info/logger/logger.go
@@ -162,7 +162,9 @@ func (l *Logger) ValidateAndUpdateLogFilepath() {
 
 func (l *Logger) Close() {
 	if l.file != nil {
-		l.file.Close()
+		if err := l.file.Close(); err != nil {
+			log.Println("Error closing log file:", err)
+		}
 	}
 	l.StopLogRotateTicker()
 }


tobychui:
Hi @PassiveLemon , I have added some new startup flags that you might also want to update the docker file to expose them to the users.

```
  -enablelog
        Enable system wide logging, set to false for writing log to STDOUT only (default true)
  -enablelogcompress
        Enable log compression for rotated log files (default true)
...
  -logrotate int
        Enable log rotation and set the maximum log file size in KB (e.g. 25 for 25KB), set to 0 for disable
```

Thanks!

AnthonyMichaelTDM:
> Hi @/PassiveLemon , I have added some new startup flags that you might also want to update the docker file to expose them to the users.
> 
> ```
>   -enablelog
>         Enable system wide logging, set to false for writing log to STDOUT only (default true)
>   -enablelogcompress
>         Enable log compression for rotated log files (default true)
> ...
>   -logrotate int
>         Enable log rotation and set the maximum log file size in KB (e.g. 25 for 25KB), set to 0 for disable
> ```
> 
> Thanks!

Since logging is enabled by default, maybe a `disablelog` and `disablelogcompress` naming convention would better communicate the meaning of the flags?

Also, I think we need to update the wiki

AnthonyMichaelTDM (commented):
I'd like to block this pending the merge, closure, or resolution of:
- #807 
- #806
- #752

PR #807, opened 2025-09-06
Update example plugins

AnthonyMichaelTDM:
Ran build_all.sh to update zoraxy_plugin for the example plugins, after addressing the issue mentioned in the commit message for 1c84a8f9cf25872c50d58f310ec475da98f32228 . Which, for posterity, is quoted below:

> [fix: remove events import from zoraxy_plugin](https://github.com/tobychui/zoraxy/commit/1c84a8f9cf25872c50d58f310ec475da98f32228)
> 
> The import, when the code is copied to develop a plugin, results an invalid path.
> 
> Fixing the path manually as a plugin developer is easy, but it shouldn't be necessary.
To fix that, the type is replaced with a string in zoraxy_plugin.IntroSpect and validation is added to lifecycle.go to ensure all subscribed events are valid.
>
> A downside is that the list of validEventNames has to be updated whenever a new event is created, but this is mitigated by placing definitions of that list and the actual event names right next to each other.

To prevent things like this from slipping through in the future, I'll add a job to #752 to automatically check that the zoraxy_plugin module is up-to-date as part of the in-progress CI system.

### DIFF
--- a/example/plugins/.gitignore
+++ b/example/plugins/.gitignore
@@ -1,0 +1,8 @@
+api-call-example/api-call-example
+debugger/debugger
+dynamic-capture-example/dynamic-capture-example
+event-subscriber-example/event-subscriber-example
+helloworld/helloworld
+restful-example/restful-example
+static-capture-example/static-capture-example
+upnp/upnp
--- a/example/plugins/api-call-example/mod/zoraxy_plugin/events/events.go
+++ b/example/plugins/api-call-example/mod/zoraxy_plugin/events/events.go
@@ -1,0 +1,133 @@
+package events
+
+import (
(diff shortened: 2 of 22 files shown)


tobychui:
@AnthonyMichaelTDM I guess you have mistakenly pushed the plugin binaries to the commit. Can you please remove them so that I can merge? Thanks!

AnthonyMichaelTDM:
Oh yeah that was a mistake, I'll fix it when I get home

AnthonyMichaelTDM:
fixed

AnthonyMichaelTDM:
>     1. More interface function to be added

I'll do this in a followup PR I plan to work on today

PR #810, opened 2025-09-07
feat(event system): Flesh out EventPayload interface

### LINKED ISSUE #806: feat(event system): Flesh out EventPayload interface
Add additional methods like `GetEventSource()` and `GetPayloadType()` to the EventPayload interface to future-proof it.

Ref: https://github.com/tobychui/zoraxy/pull/753#discussion_r2311742914_

AnthonyMichaelTDM:
closes #806

### DIFF
--- a/src/mod/eventsystem/event_system.go
+++ b/src/mod/eventsystem/event_system.go
@@ -4,8 +4,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/google/uuid"
 	"imuslab.com/zoraxy/mod/info/logger"
-	// "imuslab.com/zoraxy/mod/plugins"
 
 	"imuslab.com/zoraxy/mod/plugins/zoraxy_plugin/events"
 )
@@ -102,6 +102,7 @@ func (em *eventManager) Emit(payload events.EventPayload) error {
 	event := events.Event{
 		Name:      eventName,
 		Timestamp: time.Now().Unix(),
+		UUID:      uuid.New().String(),
 		Data:      payload,
 	}
 
--- a/src/mod/plugins/zoraxy_plugin/events/events.go
+++ b/src/mod/plugins/zoraxy_plugin/events/events.go
@@ -17,7 +17,10 @@ type EventPayload interface {
 // Event represents a system event
 type Event struct {
 	Name      EventName    `json:"name"`
-	Timestamp int64        `json:"timestamp"` // Unix timestamp
+	// Unix timestamp
+	Timestamp int64        `json:"timestamp"`
+	// UUID for the event
+	UUID      string       `json:"uuid"`
 	Data      EventPayload `json:"data"`
 }
 
@@ -88,6 +91,7 @@ func ParseEvent(jsonData []byte, event *Event) error {
 	var temp struct {
 		Name      EventName `json:"name"`
 		Timestamp int64     `json:"timestamp"`
+		UUID      string    `json:"uuid"`
 	}
 	if err := json.Unmarshal(jsonData, &temp); err != nil {
 		return err
@@ -96,6 +100,7 @@ func ParseEvent(jsonData []byte, event *Event) error {
 	// Set the event name and timestamp
 	event.Name = temp.Name
 	event.Timestamp = temp.Timestamp
+	event.UUID = temp.UUID
 
 	// Now, based on the event type, unmarshal the specific payload
 	switch temp.Name {
--- a/example/plugins/api-call-example/mod/zoraxy_plugin/events/events.go
+++ b/example/plugins/api-call-example/mod/zoraxy_plugin/events/events.go
@@ -17,7 +17,10 @@ type EventPayload interface {
 // Event represents a system event
 type Event struct {
 	Name      EventName    `json:"name"`
-	Timestamp int64        `json:"timestamp"` // Unix timestamp
+	// Unix timestamp
+	Timestamp int64        `json:"timestamp"`
+	// UUID for the event
+	UUID      string       `json:"uuid"`
 	Data      EventPayload `json:"data"`
 }
 
@@ -88,6 +91,7 @@ func ParseEvent(jsonData []byte, event *Event) error {
 	var temp struct {
 		Name      EventName `json:"name"`
 		Timestamp int64     `json:"timestamp"`
+		UUID      string    `json:"uuid"`
 	}
 	if err := json.Unmarshal(jsonData, &temp); err != nil {
 		return err
@@ -96,6 +100,7 @@ func ParseEvent(jsonData []byte, event *Event) error {
 	// Set the event name and timestamp
 	event.Name = temp.Name
 	event.Timestamp = temp.Timestamp
+	event.UUID = temp.UUID
 
 	// Now, based on the event type, unmarshal the specific payload
 	switch temp.Name {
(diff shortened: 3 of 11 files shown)


src/mod/eventsystem/event_system.go
```diff
@@ -86,31 +86,78 @@ func (em *eventManager) UnregisterSubscriber(listenerID ListenerID) error {
...
+	em.mutex.RLock()
+	defer em.mutex.RUnlock()
+	for _, listenerID := range listenerIDs {
 		listener, exists := em.subscribers[listenerID]
 
 		if !exists {
 			em.logger.PrintAndLog("event-system", "Failed to get listener for event dispatch, removing "+string(listenerID)+" from subscriptions", nil)
+			// Remove the listener from the subscription list
+			// This is done in a separate goroutine to avoid deadlock
+			go em.UnregisterSubscriber(listenerID)
```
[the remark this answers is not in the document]
AnthonyMichaelTDM: there's no deadlock, but the critical section could be shortened

src/mod/eventsystem/event_system.go
```diff
@@ -86,31 +86,78 @@ func (em *eventManager) UnregisterSubscriber(listenerID ListenerID) error {
...
+func (em *eventManager) emitTo(listenerIDs []ListenerID, event events.Event) {
+	if len(listenerIDs) == 0 {
+		return
+	}
+
+	// Dispatch to all specified listeners asynchronously
+	listenersToUnregister := []ListenerID{}
+	for _, listenerID := range listenerIDs {
+		em.mutex.RLock()
+		defer em.mutex.RUnlock()
```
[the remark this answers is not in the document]
AnthonyMichaelTDM: > defer statement will execute the unlock at the end of the function

Oh, I figured it just ran at the end of the current scope

AnthonyMichaelTDM:
This is ready for review, I have an implementation for the API endpoint needed for plugin<->plugin communication ready (patch below), I can commit it to this PR or throw it into a follow-up PR.

If you'd like, I can also create an example plugin (actually, 2 plugins) to demonstrate plugin<->plugin communication

<details>
<summary>here's the patch with my current implementation</summary>

```patch
diff --git a/src/mod/plugins/store.go b/src/mod/plugins/store.go
index fc17833..7f96bc8 100644
--- a/src/mod/plugins/store.go
+++ b/src/mod/plugins/store.go
@@ -12,7 +12,9 @@ import (
 	"strings"
 	"time"
 
+	"imuslab.com/zoraxy/mod/eventsystem"
 	"imuslab.com/zoraxy/mod/plugins/zoraxy_plugin"
+	"imuslab.com/zoraxy/mod/plugins/zoraxy_plugin/events"
 	"imuslab.com/zoraxy/mod/utils"
 )
 
@@ -354,3 +356,38 @@ func (m *Manager) HandleUninstallPlugin(w http.ResponseWriter, r *http.Request)
 
 	utils.SendOK(w)
 }
+
+// HandleEmitCustomEvent is the handler for emitting a custom event from a plugin to other plugins
+func (m *Manager) HandleEmitCustomEvent(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		utils.SendErrorResponse(w, "Method not allowed")
+		return
+	}
+
+	if r.Header.Get("Content-Type") != "application/json" {
+		utils.SendErrorResponse(w, "Invalid Content-Type")
+		return
+	}
+
+	// parse the event payload from the request body
+	var payload events.CustomEvent
+	if err := json.NewDecoder(r.Body).Decode(&payload); err != nil {
+		utils.SendErrorResponse(w, "Failed to parse event: "+err.Error())
+		return
+	}
+
+	// collect the recipients
+	if len(payload.Recipients) > 0 {
+		recipients := []eventsystem.ListenerID{}
+		for _, rid := range payload.Recipients {
+			recipients = append(recipients, eventsystem.ListenerID(rid))
+		}
+		// Emit the event to subscribers and specified recipients
+		eventsystem.Publisher.EmitToSubscribersAnd(recipients, &payload)
+	} else {
+		// Emit the event to all subscribers
+		eventsystem.Publisher.Emit(&payload)
+	}
+
+	utils.SendOK(w)
+}
diff --git a/src/plugin_api.go b/src/plugin_api.go
index 21550b5..fc197bd 100644
--- a/src/plugin_api.go
+++ b/src/plugin_api.go
@@ -108,6 +108,10 @@ func RegisterPluginRestAPI(authRouter *auth.PluginAuthMiddleware) {
 	authRouter.HandleFunc("/api/plugins/groups/list", pluginManager.HandleListPluginGroups)
 
 	authRouter.HandleFunc("/api/plugins/store/list", pluginManager.HandleListDownloadablePlugins)
+
+	// recall that these plugin APIs are under the /plugin path, so
+	// the full path to this endpoint is /plugin/event/emit
+	authRouter.HandleFunc("/event/emit", pluginManager.HandleEmitCustomEvent)
 }
 
 /* Register all the APIs */
```

</details>

src/mod/plugins/zoraxy_plugin/events/events.go
```diff
@@ -82,12 +98,31 @@ func (e *AccessRuleCreatedEvent) GetName() EventName {
...
 
+func (e *AccessRuleCreatedEvent) GetEventSource() string {
+	return "accesslist-api"
+}
+
+type CustomEvent struct {
+	SourcePlugin string         `json:"source_plugin"`
+	Recipients   []string       `json:"recipients"`
+	Payload      map[string]any `json:"payload"`
+}
```
AnthonyMichaelTDM, line 109: Idea is that CustomEvents have a source, destination, and custom payload.

At the zoraxy level, we don't need to worry about the details of what that payload looks like since presumably if plugins are sending messages to eachother they can figure out their own message format.
An alternative to the `map[string]any` would be a byte array but that's a little finicky.

CODEOWNERS
```diff
@@ -10,6 +10,8 @@
 # jemmy1794 maintains the stream proxy module
 /src/mod/streamproxy @jemmy1794
 
-# AnthonyMichaelTDM maintains the plugins module
+# AnthonyMichaelTDM maintains the plugin and event systems
 /src/mod/plugins @AnthonyMichaelTDM
 /example/plugins @AnthonyMichaelTDM
+/src/plugin_*.go @AnthonyMichaelTDM
+/src/mod/eventsystem @AnthonyMichaelTDM
```
AnthonyMichaelTDM: updating my ownership to include the event system, and other plugin-related files that weren't previously included

src/mod/eventsystem/event_system.go
```diff
@@ -86,31 +86,78 @@ func (em *eventManager) UnregisterSubscriber(listenerID ListenerID) error {
...
 		Data:      payload,
 	}
 
 	// Dispatch to all subscribers asynchronously
-	for _, listenerID := range subscribers {
+	em.emitTo(subscribers, event)
+}
+
+// Dispatch event to all specified listeners asynchronously
+func (em *eventManager) emitTo(listenerIDs []ListenerID, event events.Event) {
```
AnthonyMichaelTDM, line 145: Consolidate logic shared between Emit and EmitToSubscribersAnd

src/mod/eventsystem/event_system.go
```diff
@@ -86,31 +86,78 @@ func (em *eventManager) UnregisterSubscriber(listenerID ListenerID) error {
...
+
+	// Also emit to all subscribers of the event as usual
+	em.mutex.RLock()
+	defer em.mutex.RUnlock()
+	subscribers, exists := em.subscriptions[eventName]
+	if !exists || len(subscribers) == 0 {
+		return // No subscribers
+	}
+	em.emitTo(subscribers, event)
+}
```
AnthonyMichaelTDM, line 118: in addition to listeners subscribed to this event, also dispatch to the specified plugins.

This is the key function that will enable plugin<->plugin communication since CustomEvent events will specify their recipients.

PR #813, opened 2025-09-08
feat(plugin API): Plugin-to-plugin-comms

AnthonyMichaelTDM:
Follow up of #810 

add an endpoint to the plugin accessible API that facilitates plugin<->plugin comms via the event system.

The endpoint allows plugings to send `CustomEvent`s, which can have a specific recipiant and custom payload, allowing plugins to define their own communication protocols over this interface.
Zoraxy itself acts as a router, forwarding events between registered listeners (for now, that's just plugins, but the API is designed in such a way to enable additional kinds of event listeners in the future).

### DIFF
--- a/example/plugins/.gitignore
+++ b/example/plugins/.gitignore
@@ -3,6 +3,8 @@ debugger/debugger
 dynamic-capture-example/dynamic-capture-example
 event-subscriber-example/event-subscriber-example
 helloworld/helloworld
+plugin2plugin-comms-peer1/plugin2plugin-comms-peer1
+plugin2plugin-comms-peer2/plugin2plugin-comms-peer2
 restful-example/restful-example
 static-capture-example/static-capture-example
 upnp/upnp
--- a/example/plugins/README.md
+++ b/example/plugins/README.md
@@ -1,0 +1,10 @@
+# Example Plugins
+
+This directory contains example plugins that demonstrate how to create and use plugins with the main application. Each plugin is designed to showcase different features and capabilities of the plugin system.
+
+## Some Note-Worthy Examples
+
+- **api-call-example**: Demonstrates how plugins can make API calls to zoraxy
+- **event-subscriber-example**: Shows how to subscribe to and handle events from zoraxy within the application
+- **plugin2plugin-comms-peer1**: Illustrates communication between two plugins via the event system, where this plugin acts as the first peer
+- **plugin2plugin-comms-peer2**: Similar to the above, but this plugin acts as the second peer in the communication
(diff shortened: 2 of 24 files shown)


src/plugin_api.go
```diff
@@ -108,6 +108,10 @@ func RegisterPluginRestAPI(authRouter *auth.PluginAuthMiddleware) {
 	authRouter.HandleFunc("/api/plugins/groups/list", pluginManager.HandleListPluginGroups)
 
 	authRouter.HandleFunc("/api/plugins/store/list", pluginManager.HandleListDownloadablePlugins)
+
+	// recall that these plugin APIs are under the /plugin path, so
+	// the full path to this endpoint is /plugin/event/emit
+	authRouter.HandleFunc("/event/emit", pluginManager.HandleEmitCustomEvent)
```
tobychui, line 114: Hmm, I don't get why this is using absolute path `/event/` here. Did I missed another layer of router of something?  This auth router here should be the root router (aka all path start with the path as is without prefix), is this API suppose to be registered here?

AnthonyMichaelTDM: the name is a bit of a misnomer, this "authRouter" is really the plugin auth middleware, which places all its endpoints under the /plugin namespace so that plugin-accessible endpoints are separated from the rest of the zoraxy api.

AnthonyMichaelTDM: it's named authRouter so that the code remains visually similar to what's in `src/api.go` for easier maintenance

AnthonyMichaelTDM:
Okay, an issue has arisen: POST requests to zoraxy are protected by CSRF, and I'm not sure how plugin backends are supposed to generate valid CSRF tokens.

pluginUI -> plugin is fine
pluginUI -> zoraxy is fine (requires plugins to leak API keys to the frontend so is *not* a solution)
but plugin -> zoraxy doesn't work for endpoints protected by CSRF

I'm not sure what to do about this yet, but it's a big problem, not just for plugin comms

AnthonyMichaelTDM:
Ideally, plugins would bypass CSRF entirely since CSRF is not intended for server->server type connections, but I'm not sure how we should make that work since the CSRF middleware is global

AnthonyMichaelTDM:
Anyway, I have an example plugin (2 plugins really) that "work", but only when I disable the global CSRF middleware:
image

tobychui (approved):
LGTM, in the future I guess we should create a wrapper for event system so other plugin dev can use it more easily. Will merge it for now for v3.2.6

AnthonyMichaelTDM:
Yeah, a function to streamline making API requests from a plugin (by setting the Auth header for you) would nice too but is not needed for this release

PR #827, opened 2025-09-19
Update Sidebar CSS

Saeraphinx:
I started setting up Zoraxy earlier today and there were a few things that bugged me or looked a bit off. This PR updates the styling for the sidebar headers, the sidebar & content window spacing, and footer spacing.

Before:
image

After:
image

### DIFF
--- a/src/web/darktheme.css
+++ b/src/web/darktheme.css
@@ -336,7 +336,7 @@ body.darkTheme .ui.segment.advanceoptions {
 body.darkTheme .ui.segment{
     background-color: transparent !important;
     color: var(--text_color) !important;
-    border: 1px solid transparent !important;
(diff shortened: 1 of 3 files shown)


tobychui:
Hey @Saeraphinx 

It is so cool to see UI related PR! (You know, open source projects that optimize for UI is rare in the first place, PR that optimizing UI is even more rare!)

That sidebar design was from Zoraxy v2 which there were no RWD menu back in the days. It was best to utilize every inch of screen space and hence, the item extend all the way to the side of the window edge. 

The only issue with this PR is that we do not use main branch for dev, so I gonna need to move this to v3.2.7 branch instead, the rest looks good to me.

tobychui (approved):

PR #829, opened 2025-09-21
fix acme renew

jimmyGALLAND:
fix/feat: support SANs in TLS certificate handling renew

### DIFF
--- a/src/mod/acme/utils.go
+++ b/src/mod/acme/utils.go
@@ -126,7 +126,6 @@ func CertExpireSoon(certBytes []byte, numberOfDays int) bool {
 //   * Each label ≤ 63 characters
 //   * Only letters, digits, and hyphens
 //   * Labels do not start or end with a hyphen
-//   * Labels must have >= 2 characters
 //   * Full domain ≤ 253 characters
 // Returns an empty string if the domain is invalid.
 func NormalizeDomain(d string) (string, error) {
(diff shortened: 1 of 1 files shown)


jimmyGALLAND:
The work on this PR is now complete, and it’s ready for your review

src/mod/acme/utils.go
```diff
@@ -98,3 +116,52 @@ func CertExpireSoon(certBytes []byte, numberOfDays int) bool {
...
+	if len(d) > 253 {
+		return "", errors.New("domain exceeds 253 characters")
+	}
+
+	labels := strings.Split(d, ".")
+	for _, label := range labels {
+		if len(label) == 0 {
+			return "", errors.New("Domain '" + d + "' not valid: Empty label")
+		}
+		if len(label) < 2 {
```
zen8841: I think label == 1 is allow in RFC1035

> The labels must follow the rules for ARPANET host names.  They must
> start with a letter, end with a letter or digit, and have as interior
> characters only letters, digits, and hyphen.  There are also some
> restrictions on the length.  Labels must be 63 characters or less.
> 
> For example, the following strings identify hosts in the Internet:
> 
> A.ISI.EDU XX.LCS.MIT.EDU SRI-NIC.ARPA
>

jimmyGALLAND: You're right, my apologies. I'm committing the change.

However, I think a label can start with a digit.  "... They must start with a letter, end with a letter or digit ... " 
examples: 3com.com , 7zip.org

tobychui: 圖片

Yes, [RFC1123](https://datatracker.ietf.org/doc/html/rfc1123#section-2) have relaxed the label for allowing the first character to be a digit instead. I guess this should be ready to merge. You ok with that @zen8841 ?

zen8841: I think this is ready to merge

PR #840, opened 2025-10-09
Add authentication display to HTTP proxy list

xaxy55:
This pull request enhances the HTTP Proxy management UI by adding authentication visibility and improving menu consistency. The most important changes are:

**Authentication display in HTTP Proxy table:**

* Added a new "Authentication" column to the HTTP Proxy list table, showing the authentication method for each proxy entry with an icon and label (Basic Auth, Forward Auth, OAuth2, or None). This helps users quickly see which authentication is configured for each proxy.

### DIFF
--- a/src/web/components/httprp.html
+++ b/src/web/components/httprp.html
@@ -232,9 +232,6 @@
                     <a class="item hrpedit_menu_item" cfgpage="vdirs">
                         <i class="angle folder icon"></i> <span class="editorSideMenuText">Virtual Directory</span>
                     </a>
-                    <a class="item hrpedit_menu_item" cfgpage="vdirs">
-                        <i class="angle folder icon"></i> <span class="editorSideMenuText">Virtual Directory</span>
-                    </a>
                     <a class="item hrpedit_menu_item" cfgpage="alias">
                         <i class="at icon"></i> <span class="editorSideMenuText">Alias</span>
                     </a>


tobychui:
Hey @xaxy55 

Thanks for the PR! 

We used to have that shown in the http proxy list, but it was removed due to the interface is too cluttered and beginners are having difficulties finding the function they wanted.

Based on the history of this software evolution, I don't think adding the authentication display to HTTP proxy list again is a good idea. Do you got any particular reasons for adding this again to the HTTP proxy list?

lbr88:
Hi, new user to Zoraxy here not the PR author but just wanted to chime in.
I just migrated to Zoraxy and not having the access information after migrating 20+ services made this a bit of a pain because when going through and bulk adding everything it does not reset the form, so the access field for some internal services i setup was setup incorrectly. I just went through everything by clicking on them to verify that they were correctly configured. Having the access information would have been helpful.

If you don't want clutter maybe a compromise is a column selector that leaves it out by default or maybe a more advanced search option to select on different settings instead of just name/tag?

tobychui:
> I just went through everything by clicking on them to verify that they were correctly configured. Having the access information would have been helpful.

Ok this make sense. In fact I also hate how NPM hide settings inside multiple menus just to make it mobile friendly. Let me merge this first and implement something for showing detail views.

PR #848, opened 2025-10-14
Add support for Proxy Protocol V1 and V2 in streamproxy configuration

jemmy1794:
- Updated HTML form to allow selection of Proxy Protocol version.
- Backend and config struct now use ProxyProtocolVersion (int) instead of UseProxyProtocol (bool).
- UI and API now pass and display Proxy Protocol version.

TODO: UDP should only allow Proxy Protocol V2


tobychui (approved):
LGTM

PR #850, opened 2025-10-16
Add Uptime Monitor checkbox to new rule page.

Ayandaftary:
I used the edit function and the implementation in http rules page as guide to copy it to the add function. It adds the option beside the Sticky session checkbox.


Ayandaftary:
Completely missed the existing commit to add this function

Issue #867, opened 2025-10-29
[ENHANCEMENTS] Adding Joohoi’s ACME-DNS.


axiades:
**Is your feature request related to a problem? Please describe.**
I am Using seperately for my Domains hosted at denic, the Acme-DNS Challenge Server but the Support for this is not implemented in Zoraxy

**Describe the solution you'd like**
[Joohoi’s ACME-DNS](https://github.com/joohoi/acme-dns)
https://go-acme.github.io/lego/dns/acme-dns/index.html

zen8841:
@axiades Can you test the acmedns(janhoi acme-dns) provider work properly?

https://github.com/zen8841/zoraxy/tree/updatelego

axiades:
How can i update if i have installed Zoraxy through 
```
https://community-scripts.github.io/ProxmoxVE/scripts?id=zoraxy
bash -c "$(curl -fsSL https://raw.githubusercontent.com/community-scripts/ProxmoxVE/main/ct/zoraxy.sh)"
 ```
this

**Comment by zen8841** [comment]
If you don't want to compile zoraxy from source.
You can wait for zoraxy v3.2.9.
https://github.com/tobychui/zoraxy/pull/870
v3.2.9 is ready.
Please report the test result after update.
Thank you very much.

**Comment by axiades** [comment]
I Have Updated Zoraxy.
But im too Dumb for Using Acme-DNS..

With Acme.sh Script im Using the AcmeDNS Server follwing Way:
export ACMEDNS_UPDATE_URL=""
export ACMEDNS_USERNAME=""
export ACMEDNS_PASSWORD=""
export ACMEDNS_SUBDOMAIN=""

But now i Need an APIBase which is the AcmeDNS-Update-URL, an StorageBaseURL which i Dont have, and the json File.
i Get only an Error that the NameServer of AcmeDNS did not Retunn the Expected TXT Record

**Comment by zen8841** [comment]
According to [lego docs](https://go-acme.github.io/lego/dns/acme-dns/index.html), STORAGE_PATH and STORAGE_BASE_URL can only choose one to use.
Also in https://github.com/go-acme/lego/blob/master/providers/dns/acmedns/acmedns.go .
```go
	if config.StoragePath != "" && config.StorageBaseURL != "" {
		return nil, errors.New("storagePath and storageBaseURL cannot be used at the same time")
	}
```

I read the lego source code. I think the Json structure is like this.
```json
{
[...]

Issue #866, opened 2025-10-29
[ENHANCEMENTS] Add KeyHelp DNS Challenge


axiades:
**Is your feature request related to a problem? Please describe.**
i use Keyhelp for domain Managing and want tu Pull Wildcard Certificates for my Domains managed in Keyhelp with Zoraxy.

Sorry my English is bad..

**Describe the solution you'd like**
Adding the [KeyHelp DNS](https://go-acme.github.io/lego/dns/keyhelp/index.html) Challenge from Go-Acme

zen8841:
@axiades Can you test the keyhelp provider work properly?

https://github.com/zen8841/zoraxy/tree/updatelego

axiades:
KeyHelp is Working! Nice!!

PR #869, opened 2025-10-30
Update lego to v4.27.0

zen8841:
Update lego to v4.27.0

Add dns challenge providers:
acmedns(janhoi acme-dns) #867
keyhelp #866
others dns providers

Need test for acmedns and keyhelp providers.
But these change won't affect other function, I think it's safe to merge.


tobychui (approved):
Nice work! You even updated the code gen for the acmedns!

PR #883, opened 2025-11-08
Optimize Github Action for creating the Docker image

dwydler:
This pull request enhances the Github Action "Build and push Docker image". The most important changes are:
- Sec: The permissions section has been added ([Link](https://docs.github.com/en/actions/reference/security/secure-use#use-secrets-for-sensitive-information)).
- Enc: The caching method has been switched from gha to registry cache. Reduce the build time of new versions approximately 55 minutes.
- Sec: Pin actions to a full-length commit SHA ([Link](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions)).
- Enc: The file schema was checked and corrected using yamllint.
- Enc:The tag detection for the Docker image has been revised.
- Enc: Static values ​​outsourced to ENV variable.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -47,7 +47,6 @@ jobs:
           else
             echo "docker=${{ env.DOCKER_ACCOUNT }}/${{ env.DOCKER_REPO }}:latest, ${{ env.DOCKER_ACCOUNT }}/${{ env.DOCKER_REPO }}:${{ github.event.release.tag_name }}" >> $GITHUB_OUTPUT
           fi
-          cache-to: type=registry,ref=${{ env.DOCKER_ACCOUNT }}/${{ env.DOCKER_REPO }}:${{ env.DOCKER_REPO_TAG_CACHE_NAME }},mode=max
 
       - name: Build and push Docker image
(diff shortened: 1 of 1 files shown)


tobychui:
@dwydler Thanks for the PR!
I will forward the review request to our docker maintainer who also build the Github action pipeline.

.github/workflows/docker.yml
```diff
@@ -1,60 +1,59 @@
+---
 name: Build and push Docker image
 
-on:
+"on":
```
PassiveLemon, line 4: Why is "on" in quotes, and what's with the 3 leading dashes?

PassiveLemon: other than that, it lgtm

dwydler: The three leading dashes come from the YAML 1.2 specification, section Document Markers.

> Why is "on" in quotes

In older YAML versions (e.g., YAML 1.1), the word `on` was interpreted as a Boolean value, like `true` or `false`. Quotation marks are used to clearly indicate that `on` is a string (i.e., a key name). Therefore, "on" is used for compatibility and security.

PassiveLemon: seems alright then, thanks

tobychui: Cool, so this is ready to merge?

PR #893, opened 2025-11-15
Add tzdata package and timezone environment variable to Dockerfile

DeusMaximus:
This resolves an issue where the time is set to UTC, regardless of sharing /etc/localtime or /etc/timezone from the host.

For instance, being from Australia (GMT +10), I constantly get the error that the date is set in the future when attempting to look at logs within Zoraxy.


PassiveLemon:
Does this mean that there is no longer a need to share /etc/localtile? If so, would you mind removing mentions of it from the Docker section Readme? other than that, lgtm

PassiveLemon:
Perfect, thank you!

PR #942, opened 2025-12-23
Fix incorrect status codes and missing error pages in production

Restorn:
Replaces `http.ServeFile` usage for serving error pages to resolve issues in the production build.

- `ServeFile` attempts to read from the local disk. While these files exist during development, they are not included in the production binary, causing the error handlers to fail.
- `http.ServeFile` automatically sets the status code to 200 OK on success. When it failed to find the file (as noted above), it implicitly returned 404 Not Found, overriding the intended error codes.

### DIFF
--- a/src/web/forbidden.html
+++ b/src/web/forbidden.html
@@ -1,0 +1,55 @@
+<html>
+    <head>
+        <!-- Zoraxy Forbidden Template -->
+        <meta charset="UTF-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0 user-scalable=no">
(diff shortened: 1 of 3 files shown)


tobychui:
Thanks for the PR!

I see you have deleted the two error page html files from the `src/web/` folder. However, the `src/web/` is the correct location for them and in zoraxy implementation, those are fallback files to serve if the same file cannot be found under the templates folder. 

i.e. The error page is embedded into the binary, but the user can optionally place a error page file in the `www/templates/` folder to override them if they wanted to. 

That is why this pull request technically fixed a bug but this is not using a correct implementation which most Zoraxy user are expected. You mind changing the implementation a bit to fix that? Or I would close this and implement my own fix on this?

tobychui (approved):
Ok, there are still something off but I will merge it and fix it on my branch. Thanks for the PR :)

PR #943, opened 2025-12-24
UPDATE_GEOIP should only update if true

kuhnchris:
Passing "UPDATE_GEOIP" to the executable in docker seems like a pretty weird idea, as that leads to updating the update again and exiting.
I assume the intended use for this was, to only run "update_geoip" if it is set (in the main function) and simply omit it in the run function.

If this was indeed not the intended use (i.e. allowing for 2 calls of `zoraxy -update_geoip=true` during the entrypoint.py) then please let me know.

Thanks

### DIFF
--- a/docker/entrypoint.py
+++ b/docker/entrypoint.py
@@ -100,7 +100,6 @@ def start_zoraxy():
     f"-plugin={ getenv('PLUGIN', '/opt/zoraxy/plugin/') }",
     f"-port=:{ getenv('PORT', '8000') }",
     f"-sshlb={ getenv('SSHLB', 'false') }",
-    f"-update_geoip={ getenv('UPDATE_GEOIP', 'false') }",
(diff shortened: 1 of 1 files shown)


PassiveLemon:
Oh that's a good catch. I think a better solution would be to keep only the Zoraxy flag, but default to true and remove the if statement portion

kuhnchris:
Well, my particular use-case was an air-gapped environment that needs to be started with docker - so basically the start kept hanging on the "update_geoip" part until it was done despite having geoip disabled.
Just removing the "if" portion and keeping the zoraxy flag would mean that the container would start once, do the update_geoip path, and then "fail" as it exists (the update_geoip specifically only does the updating and then exits), so I guess we would need to keep the "if" part if we want to give the possibility to do update_geoip with running the docker container. Else you'd need 2 containers, one for update and one for running, no?

tobychui (approved):

PR #944, opened 2025-12-24
[docs] Fix typo in SniffResult constant

kuhnchris:
Just a quick typo in the docs. (not sure we should mention to also include "strings"? but this is a plugin doc, not a go beginners tutorial)


tobychui (approved):

tobychui:
> not sure we should mention to also include "strings"?

Not necessary, the type of a typed constant is the constant itself, so the abstracted type of SniffResult is `plugin.SniffResult`.

PR #945, opened 2025-12-25
fail guard port 80 not available

kuhnchris:
instead of panicing and crashing, we should allow the server to run without explicit HTTP->HTTPS forwarding. 
additionally I'd even propose we should add the http->https forwarding as a switch.
and actually, i think it would make more sense to also make the http port configurable, especially if you are using this in combination with docker. 

But for starters this just guards against panicing. Please advice if you'd also like to have the aforementioned configurability and/or switches.

There is still a case where the ACME part will forcefully try to start a port 80 server, but i couldnt reproduce it right now, so it may also affect that.

### DIFF
--- a/docker/entrypoint.py
+++ b/docker/entrypoint.py
@@ -100,7 +100,6 @@ def start_zoraxy():
     f"-plugin={ getenv('PLUGIN', '/opt/zoraxy/plugin/') }",
     f"-port=:{ getenv('PORT', '8000') }",
     f"-sshlb={ getenv('SSHLB', 'false') }",
-    f"-update_geoip={ getenv('UPDATE_GEOIP', 'false') }",
     f"-version={ getenv('VERSION', 'false') }",
     f"-webfm={ getenv('WEBFM', 'true') }",
     f"-webroot={ getenv('WEBROOT', './www') }",
@@ -115,8 +114,9 @@ def main():
   print("Updating CA certificates...")
   run(["update-ca-certificates"])
 
-  print("Updating GeoIP data...")
-  run(["zoraxy", "-update_geoip=true"])
+  if getenv("UPDATE_GEOIP", "false").lower() == "true":
+    print("Updating GeoIP data...")
+    run(["zoraxy", "-update_geoip=true"])
 
   os.chdir("/opt/zoraxy/config/")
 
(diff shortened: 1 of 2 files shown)


tobychui:
Thanks for raising this issue. There is already a check right here, but just placed at an incorrect location.
```go
	listenOnPort80 := true
	if netutils.CheckIfPortOccupied(80) {
		listenOnPort80 = false
	}
```

So instead of this
```go
    listenOnPort80 := true
	if netutils.CheckIfPortOccupied(80) {
		listenOnPort80 = false
	}
	sysdb.Read("settings", "listenP80", &listenOnPort80)
	if listenOnPort80 {
		SystemWideLogger.Println("Port 80 listener enabled")
	} else {
		SystemWideLogger.Println("Port 80 listener disabled")
	}

```

This can be easily fixed by swapping the config priority like this
```go
	listenOnPort80 := true
	sysdb.Read("settings", "listenP80", &listenOnPort80)
	if netutils.CheckIfPortOccupied(80) {
		listenOnPort80 = false
	}
	if listenOnPort80 {
		SystemWideLogger.Println("Port 80 listener enabled")
	} else {
		SystemWideLogger.Println("Port 80 listener disabled")
	}
```

This avoids adding another variable and making the logic more complex. I will closing this and swap it on my dev branch real quick.

tobychui:
Ok this involve a much deeper problem then just a single switch on port 80 listener, especially after adding the secondary listener port designs. Let me think of a better way to fix it.

kuhnchris:
> Thanks for raising this issue. There is already a check right here, but just placed at an incorrect location.
> 
> ```go
> 	listenOnPort80 := true
> 	if netutils.CheckIfPortOccupied(80) {
> 		listenOnPort80 = false
> 	}
> ```
> 
> So instead of this
> 
> ```go
>     listenOnPort80 := true
> 	if netutils.CheckIfPortOccupied(80) {
> 		listenOnPort80 = false
> 	}
> 	sysdb.Read("settings", "listenP80", &listenOnPort80)
> 	if listenOnPort80 {
> 		SystemWideLogger.Println("Port 80 listener enabled")
> 	} else {
> 		SystemWideLogger.Println("Port 80 listener disabled")
> 	}
> ```
> 
> This can be easily fixed by swapping the config priority like this
> 
> ```go
> 	listenOnPort80 := true
> 	sysdb.Read("settings", "listenP80", &listenOnPort80)
> 	if netutils.CheckIfPortOccupied(80) {
> 		listenOnPort80 = false
> 	}
> 	if listenOnPort80 {
> 		SystemWideLogger.Println("Port 80 listener enabled")
> 	} else {
> 		SystemWideLogger.Println("Port 80 listener disabled")
> 	}
> ```
> 
> This avoids adding another variable and making the logic more complex. I will closing this and swap it on my dev branch real quick.

yeah, but sadly further down there is also a flip for the switch back on when doing the Force HTTP->HTTPS. As you mentioned at the end that's a bigger topic I guess. Thanks for taking a look at it.

PR #948, opened 2025-12-26
allow plugin API to modify proxy upstreams

kuhnchris:
Allow Plugin API to modify the /proxy/upstream/* path.

Would make writing plugins that modify the upstream based on conditions a lot easier. :-)


tobychui:
Cool, but please be aware that I do not guarantee backward compatibility of the post payload for proxy add endpoint. So it would actually be better (architecturally) to have a wrapper in the plugin module for keeping the plugin api "stable". 

I will merge this for now.

kuhnchris:
> Cool, but please be aware that I do not guarantee backward compatibility of the post payload for proxy add endpoint. So it would actually be better (architecturally) to have a wrapper in the plugin module for keeping the plugin api "stable".
> 
> I will merge this for now.

Definitly, maybe even a socket communication instead of using the regular frontend API, as if we turn off the administrative interface we probably also lose the API functionality (not tested, but from what I saw so far).

Additionally (i started that with the help of AI, to tell the truth) we should provide the return values of the api calls in the zoraxy_plugin aswell, else everyone gonna reinvent the wheel when trying to parse the replies.

PR #1045, opened 2026-02-16
fix: forward auth saving and defaults

CrazyWolf13:
Hi,

As noted here: #895 a PR was prefered, I saw noone has yet created a PR so I created this PR.

This adresses an issue, where ReponseHeaders were being written to the client instead of the server.

This should also adress an issue, where `strings.Split` was used instead of `cleanSplit`, which caused empty fields to be stored as `[""]` instead of `nil`, breaking the "copy all if empty" default behavior after saving settings.
The same is applied already above on line 72 to 76 for example: https://github.com/CrazyWolf13/zoraxy/blob/3fd7474a16eb4e3a66b242ee3905a8f58df29893/src/mod/auth/sso/forward/forward.go#L72-L76


james-d-elliott:
Thanks, I've been flat out!

PR #1097, opened 2026-03-30
feat: better plugin manager

kjagosz:
This PR improves the plugin manager:

CleanShot 2026-03-30 at 20 56 10

Changes:
- Add advanced settings for plugin store URLs, including persistence of changes
- Refactor plugin introspection and store utilities to use absolute paths
- Introduce a handler for retrieving and saving plugin store URLs
- Update plugin list synchronization to support multiple sources and default URLs
- Improve the UI for managing plugin store repositories with real-time updates

### DIFF
--- a/src/mod/plugins/store.go
+++ b/src/mod/plugins/store.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	urlpkg "net/url"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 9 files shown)


src/mod/plugins/utils.go
```diff
@@ -32,6 +32,25 @@ func (m *Manager) GetPluginEntryPoint(folderpath string) (string, error) {
...
+			if _, err := os.Stat(startBatchPath); err == nil {
+				return startBatchPath, nil
+			}
+		} else {
+			if _, err := os.Stat(startScriptPath); err == nil {
+				return startScriptPath, nil
+			}
+		}
+	}
+
```
AnthonyMichaelTDM, line 53: I'm not sure about this. Typically plugins *should* ship the executable, and if you're developing a plugin you can just recompile it to get the latest version.
Also, not every plugin is implemented in Go (ex: [rust api](https://github.com/aroz-online/zoraxy-rs)) so this isn't a reliable detection mechanism

AnthonyMichaelTDM (changes requested):
Other than that comment, looks good to me

src/mod/plugins/utils.go
kjagosz: You're right — this is specific to Go, so checking for go.mod here isn't necessary. We'll still get support for Windows and other operating systems.

src/mod/plugins/store.go
```diff
@@ -72,16 +150,42 @@ func (m *Manager) getPluginListFromURL(url string) ([]*DownloadablePlugin, error
 		return nil, fmt.Errorf("failed to unmarshal plugin list from %s: %w", url, err)
 	}
 
+	baseURL, _ := urlpkg.Parse(url)
+
 	// Filter out if IconPath is empty string, set it to "/img/plugin_icon.png"
 	for _, plugin := range pluginList {
```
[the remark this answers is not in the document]
AnthonyMichaelTDM: This is a good point, maybe it'd be better to pull each list in parallel with a timeout (say, 20 seconds or something) and collect results over a channel?

AnthonyMichaelTDM (commented):
Can this commit (0ca6d68f643120e2fb46e22cb5c3c9dbb801a8a2) be in a separate PR? Or is it necessary for these plugin store improvements?

PR #1160, opened 2026-05-02
fix(streamproxy): re-init atomic counters when loading rules (#1157)

### LINKED ISSUE #1157: [BUG] Zoraxy crashes with SIGSEGV when using TCP Stream Proxy (connCopy nil pointer dereference)
## Describe the bug
Zoraxy crashes with a segmentation fault (SIGSEGV) when a TCP Stream Proxy rule is enabled and receives traffic.

The crash happens inside the stream proxy module (`connCopy`) and causes the service to restart in a loop.

---

## To Reproduce

1. Create a Stream Proxy rule:
   - Name: `Drive 6690`
   - Listening Address: `0.0.0.0:6690`
   - Target: `192.168.10.5:6690`
   - Protocol: TCP
2. Set `"Running": true`
3. Restart Zoraxy
4. Connect to the port:
   nc 192.168.10.130 6690
5. Observe crash in logs:
   journalctl -u zoraxy -f

---

## Expected behavior
Zoraxy should proxy TCP traffic without crashing.

---

## Actual behavior
Zoraxy crashes with:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1e76294]

imuslab.com/zoraxy/mod/streamproxy.(*ProxyRelayInstance).connCopy
    tcpprox.go:41

Systemd restarts the service repeatedly.

---

## Screenshots
N/A (log-based issue)

---

## Browser (if it is a bug appears on the UI section of the system):
N/A

---

## Host Environment (please complete following information, DO NOT REMOVE ANY FIELD(S)):

- Arch: amd64
- Device: Proxmox LXC container
- OS: Debian GNU/Linux 13 (trixie)
- Version: Debian 13
- Are you using Docker? no
- Docker Version: N/A
- Zoraxy Version: unknown (binary installed manually, command `zoraxy --version` not available)

---

## Supplementary links
N/A

---

## Additional context

- Issue occurs only when Stream Proxy is enabled and receives traffic
- Reverse proxy (HTTP/TLS) works correctly
- Disabling the stream proxy rule ("Running": false) prevents crash
- Crash happens consistently after reboot if rule is enabled
- Stack trace always points to:
  imuslab.com/zoraxy/mod/streamproxy/tcpprox.go:41

---

## Workaround

Using HAProxy instead works correctly for the same backend:

frontend synology_drive_6690
    bind 0.0.0.0:6690
    default_backend synology_drive_backend

backend synology_drive_backend
    server synology 192.168.10.5:6690 check

This suggests the issue is specific to Zoraxy's TCP stream proxy implementation.

---

## Logs

panic: runtime error: invalid memory address or nil pointer dereference
imuslab.com/zoraxy/mod/streamproxy.(*ProxyRelayInstance).connCopy

---

## Additional diagnostic info

uname -a:
Linux zoraxy 6.17.13-4-pve #1 SMP PREEMPT_DYNAMIC PMX 6.17.13-4 (2026-04-21T22:03Z) x86_64 GNU/Linux

SAY-5:
Resumed stream-proxy rules panic in `connCopy` (`tcpprox.go:41`) on the first forwarded byte because their `*atomic.Int64` accumulators are nil after JSON load — the fields are unexported and so are skipped by json (un)marshal, while `NewConfig` initialises them on the create path only.

Allocate fresh `atomic.Int64` values for both counters right after `json.Unmarshal` succeeds, mirroring the create path. Fixes #1157.

### DIFF
--- a/src/mod/streamproxy/streamproxy.go
+++ b/src/mod/streamproxy/streamproxy.go
@@ -125,6 +125,14 @@ func NewStreamProxy(options *Options) (*Manager, error) {
 			continue
 		}
 
+		//Re-initialize unexported runtime fields. These are skipped by json
+		//(un)marshal and would otherwise be nil for resumed rules, causing a
+		//SIGSEGV in connCopy on the first forwarded byte (issue #1157).
+		aAcc := atomic.Int64{}
+		bAcc := atomic.Int64{}
+		thisRelayConfig.aTobAccumulatedByteTransfer = &aAcc
+		thisRelayConfig.bToaAccumulatedByteTransfer = &bAcc
+
 		//Append the config to the list
 		previousRules = append(previousRules, thisRelayConfig)
 	}


tobychui:
Hey @SAY-5 

Thanks for the PR! I guess this has been patched in v3.3.3 dev branch

https://github.com/tobychui/zoraxy/commit/ef3d78204a0a95fef1c6941f8eaa35c55b0bba2a

Does this work for you? Or is it not working on your side that require additional patching?

SAY-5:
Yep — ef3d782 covers the same root cause (the unexported `*atomic.Int64` not being reinstated on load). Closing in favour of your patch.

PR #1163, opened 2026-05-06
Add OpenBSD build support

byteonabeach:
Building Zoraxy on OpenBSD 7.8 (amd64) fails because the sshprox package uses symbols UseWinTTY and gotty that are only defined in files with build tags that exclude OpenBSD.

The error is:
  mod/sshprox/sshprox.go:67:5: undefined: UseWinTTY
  mod/sshprox/sshprox.go:109:5: undefined: UseWinTTY
  mod/sshprox/sshprox.go:148:20: undefined: gotty 
  mod/sshprox/sshprox.go:155:22: undefined: gotty
  mod/sshprox/sshprox.go:209:5: undefined: UseWinTTY
  mod/sshprox/utils.go:24:5: undefined: UseWinTTY
  mod/sshprox/utils.go:36:12: undefined: gotty

I add a single file, src/mod/sshprox/sshprox_openbsd.go, with the build constraint //go:build openbsd. It defines `UseWinTTY = false` and provides an embedded filesystem for the gotty directory. No existing files are modified.

After the change the package compiles and the binary links on OpenBSD. The gotty executable for OpenBSD is not embedded, so the SSH proxy feature will still return "platform not supported" at runtime. This PR only addresses the build breakage.

### DIFF
--- a/src/mod/sshprox/sshprox_openbsd.go
+++ b/src/mod/sshprox/sshprox_openbsd.go
@@ -1,0 +1,10 @@
+//go:build openbsd
+
+package sshprox
+
+import "embed"
+
+const UseWinTTY = false
+
+//go:embed gotty/*
+var gotty embed.FS


tobychui:
Hey @byteonabeach 

> This PR only addresses the build breakage.

Yes, this is exactly what I want to say.  The better way to do this is just set UseWinTTY to true and use the Zoraxy internal tty terminal feature instead. Maybe you can reference `sshprox/embed_windows_amd64.go` and modify the OpenBSD version accordingly?

tobychui (changes requested):
As previous comment, please use the internal WinTTY feature instead.

tobychui:
Well since OpenBSD is not in our supported list and this only solved compile issue (function still unusable after this patch), I think I better close this. Currently there is no plan to support OpenBSD officially, but feel free to maintain a community port for this if needed.

PR #1181, opened 2026-05-19
fix: migrate ACME DNS credentials when setting fallback certificate

### LINKED ISSUE #1124: [ACME] DNS Provider and DNS Credential configuration required for ACME Provider
**Describe the bug**
When using a (in my case cloudflare) DNS challenge in the acme tool, zoraxy seems to not save the dns credentials. In my case the auth token. So a cert renewal is impossible, neither automatically nor manually as i have to put the token in each time i want to obtain a cert. This is not user friendly, as specifically cloudflare only shows the token once. 

**Request Details**
- ACME mode: DNS challenge
- CA used: Let's Encrypt
- DNS provider: Cloudflare

**Error Message**
DNS Provider and DNS Credential configuration required for ACME Provider (Error -2)

**To Reproduce**
1. Go to TLS / SSL Certificates
2. Choose a certificate that has been created with cloudflare dns challenge
3. Click on "Renew Certificate"

**Expected Common Name (CN) in certy**
*.example.com

**Additional context**
N/A

** Host Environment (please complete following information, DO NOT REMOVE ANY FIELD(S)) **
- Arch: x86_64
- Device: Ugreen NAS
- OS: Linux
- Version latest
- Are you using Docker? yes
- Docker Version N/A
- Zoraxy Version 3.3.2

mgrusso:
## Problem

When a user clicks **Set as Default / Fallback Certificate**, the cert files are
renamed from `{hostname}.{pem,key,json}` to `default.{pem,key,json}`. However
the ACME database entries for DNS challenge credentials remain keyed by the
original hostname:

```
{hostname}_dns_provider
{hostname}_dns_credentials
{hostname}_dns_servers
```

During auto-renewal (and manual renewal via **Renew Now**), `renewExpiredDomains`
derives the certificate name from the on-disk filename — which is now `default`
— and looks up `default_dns_provider` / `default_dns_credentials` in the database.
Those keys do not exist, so the renewal fails.

Closes #1124

## Solution

- Add `ACMEHandler.MigrateACMEDNSConfig(oldName, newName string)` which copies
  `_dns_provider`, `_dns_credentials`, and `_dns_servers` database keys from
  `oldName` to `newName` and removes the originals.
- Call it inside `tlscert.Manager.SetCertAsDefault` at two points:
  1. When the **previous** default cert is restored to its original filename
     (`"default"` → original name), so its credentials remain findable.
  2. When the **new** cert is renamed to `default` (`certname` → `"default"`),
     so renewal can look them up by the new filename.
- Store an `*acme.ACMEHandler` reference on `tlscert.Manager` and wire it in
  `start.go` after both components are initialised.

## Test plan

- [ ] Configure a certificate with ACME DNS challenge credentials (provider + token).
- [ ] Click **Set as Default** for that certificate.
- [ ] Verify the cert files are renamed to `default.*` as before.
- [ ] Trigger **Renew Now** — renewal should succeed using the migrated
      `default_dns_provider` / `default_dns_credentials` keys.
- [ ] Set a different certificate as default; verify the first cert is restored
      with its original credentials intact and renewal works for it too.
- [ ] Certificates without DNS credentials (HTTP-01 challenge) are unaffected.

### DIFF
--- a/src/mod/acme/acme.go
+++ b/src/mod/acme/acme.go
@@ -102,6 +102,29 @@ func (a *ACMEHandler) Logf(message string, err error) {
 	a.Logger.PrintAndLog("ACME", message, err)
 }
 
+// MigrateACMEDNSConfig moves the DNS credential database entries stored under
+// oldName to newName, then removes the old keys. This must be called whenever a
+// certificate file is renamed (e.g. when it is promoted to or demoted from the
+// fallback/default certificate) so that auto-renewal can still locate the
+// correct DNS challenge credentials by the new filename.
+func (a *ACMEHandler) MigrateACMEDNSConfig(oldName, newName string) {
+	if !a.Database.TableExists("acme") {
+		return
+	}
+	for _, suffix := range []string{"_dns_provider", "_dns_credentials", "_dns_servers"} {
+		oldKey := oldName + suffix
+		newKey := newName + suffix
+		if !a.Database.KeyExists("acme", oldKey) {
+			continue
+		}
+		var value string
+		if err := a.Database.Read("acme", oldKey, &value); err == nil {
+			a.Database.Write("acme", newKey, value)
+			a.Database.Delete("acme", oldKey)
+		}
+	}
+}
+
 // Close closes the ACMEHandler.
 // ACME Handler does not need to close anything
 // Function defined for future compatibility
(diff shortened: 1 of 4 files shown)


tobychui (commented):
Looks like a promising fix you got here! It looks all good on my side but I didn't got the time to test it, maybe I will let @zen8841 to decide if this should merged or not.

Caros2017:
@zen8841 are you able to review this? I am currently facing the same issue. Glad @mgrusso took the time for this fix, thanks for that!

zen8841:
I've reviewed it; this method adds too much complexity, so I won't use it. I will fix this issue in version 3.3.4.

tobychui:
Hi @zen8841 

Are you still working on this recently? If no maybe I can close it for you?

zen8841:
I'm still working on it, and I'll submit PR by the end of this weekend at the latest.

Issue #1007, opened 2026-01-20
[HELP] High RAM usage using Docker


Frankynov:
**What happened?**
Hello !
I have installed Zoraxy on two different hosts. The first one in an LXC container using the proxmox community scripts, all is fine.
The second one, is on a hosted VPS running ARM Ampere CPU, with both a public ip and a private network.
I have noticed with docker stats that there is quite a big amount of RAM being used, despite having disabled the fast geoip option.

**Describe what have you tried**
Here is the docker compose I have used : 

```
services:
  zoraxy:
    image: zoraxydocker/zoraxy:latest
    container_name: zoraxy
    restart: unless-stopped
    ports:
      - 80:80
      - 443:443
      - 81:8000
    volumes:
      - /root/zoraxy/config/:/opt/zoraxy/config/
      - /root/zoraxy/plugin/:/opt/zoraxy/plugin/
      - /var/run/docker.sock:/var/run/docker.sock
#    extra_hosts:
#      - "host.docker.internal:host-gateway"
    environment:
      FASTGEOIP: "false"
      TZ: "Europe/Prague"
```

And this is what the docker stats is giving : 

```
CONTAINER ID   NAME                                        CPU %     MEM USAGE / LIMIT     MEM %     NET I/O           BLOCK I/O         PIDS 
a55cbde51cbf   zoraxy                                      0.05%     1.474GiB / 7.559GiB   19.50%    4.39MB / 5.19MB   5.31MB / 233kB    11 
```

**Describe the networking setup you are using**
- Are you using the docker build of Zoraxy? yes (see docker compose above)
- Your Zoraxy version? 3.3.0
- Are you using Cloudflare? no
- Are your system hosted under a NAT router? Yes, my services are running on private network 10.10.0.0/24, Zoraxy being on 10.10.0.2:81 instead of 8000 (8000 being already used by another service)

**Additional context**
I initially had the fastgeoip set to true, therefore I put it to false and restarted both the container and the server, without any change unfortunately.

If there are more debug options or commands I can provide let me know !

Thank you

baskinsy:
Ι have the same issue since I started to use zoraxy. Any amount of memory given to the system is consumed after some time and I have to restart zoraxy container to recover it. I haven't been able to pinpoint the issue but always happens with any configuration when running with docker. 

My setup is on a KVM VM (proxmox) with public and private IP also, on x86 host, so it seems the CPU is not the issue. 

[...]

Issue #816, opened 2025-09-10
[BUG] Memory overflow with false forward-auth redirect (Authelia)


FirebladeBMW:
**Describe the bug**
Zoraxy consumed the complete RAM of the Docker LXC (8GB) running in Proxmox and basically making Docker unresposive.
The issue is not related to FASTGEOIP

The culprit what is causing this behavior seems to be a wrong setup of AUTHELIA forward auth.

Running HTOP in the docker host shows that memory consumption of Zoraxy is raising rapidly. If not stopping Zoraxy manually it would consume the full memory within a couple minutes. (see screenshot)

When this occurs the browser window keeps loading and will not end.
Closing the browser window will not stop memory increase.

It can be stopped by 
a) shutting down the container
b) changing entries to something that will not cause a loop and reopening the host (not 100% sure but this seemed to have stopped it)

Authelia log show such entries when this occurs:

[...]

PR #1191, opened 2026-05-26
Fix/statistic bounded maps

ElmoViggiani:
Hello,
I fixed an unbounded growth bug in the statistics module with help from Claude.

"per-day statistics maps grow without bound under high-cardinality request inputs, leading to OOM-kill"

Memory usage will grow throughout the day, eventually leading to an out of memory process kill.

I'm getting millions of requests per day from many hundreds of thousands unique ip's.

The attached screenshot from the Resource Monitor plugin shows memory growing steadily before the patch, and a flat line after the patch. I'm running with -fastgeoip otherwise the cpu would be pegged.
Screenshot 2026-05-23 at 08 31 20

Before the patch, sys.db would grow 800mb+ per 24 hours and I was having to delete it at midnight and start over.
After the patch, sys.db grows much slower. Only about 160mb after 4-5 days, because there are fewer items saved to the db. I've tested -stats_max_entries with 20000, 50000, and 100000, and the memory usage has consistently stayed flat. Sys.db growth scaled with the -stats_max_entries cap, as expected.

By Claude:
In mod/statistic, the per-day DailySummary keeps a separate map for each request dimension it tracks — client IP, User-Agent, Referer, request URL, downstream and upstream hostnames, and forwarding type. Each map grows by one entry per distinct value seen, and the maps reset at midnight. For the four dimensions whose values come straight from the request stream (client IP, User-Agent, Referer, URL), the set of possible values is effectively unbounded — every distinct value lands as a new permanent map entry until the next reset. On a busy proxy this can drive memory consumption well into the gigabytes over a day of uptime, amplified by the autosave path which re-serializes the whole summary to disk every ten minutes. A handful of past issues have reported Zoraxy gradually consuming all available memory across multi-day cycles (most notably #1007 and parts of the discussion on #816); the existing -fastgeoip workaround addresses a different startup-time allocation and doesn't help with this runtime growth.

This patch adds a per-map size cap that's opt-in via a new CLI flag, -stats_max_entries=N. When unset (or =0), behavior is byte-for-byte identical to the current code — same Load → check → Store pattern, no overhead, no risk to existing deployments. When set to a positive value, each of the eight maps stays at or under N entries. Once a map fills up, the next new insert triggers a quick trim: it scans the map, sorts entries by how many requests they've received, and drops the lowest-count 10% to make room. Existing entries continue to count up at the same speed they always did; the trim only runs on the relatively rare event of a new entry arriving at a full map, and trims that happen at the same time are serialized so request handling never blocks on stat collection.

The choice to drop the lowest-count entries (rather than oldest, newest, or random) means the data the dashboards actually display is preserved. The top-IPs, top-User-Agents, top-URLs panels all surface the most-frequently-seen items in each dimension — and those are exactly the entries the cap keeps. What gets dropped is the long tail: a single visit from an unusual IP, a malformed Referer, a one-time URL request. None of those would have shown up on the dashboard anyway. The top-line totals (TotalRequest / ValidRequest / ErrorRequest) are completely untouched, so the "X requests today" headline remains exact regardless of the cap. Recommended starting value is -stats_max_entries=20000, which gives plenty of headroom for legitimate diversity while bounding the stats structure at roughly 32 MB worst case across all eight maps.

****

### DIFF
--- a/src/def.go
+++ b/src/def.go
@@ -84,6 +84,9 @@ var (
 	/* Logging Configuration Flags */
 	enableLog = flag.Bool("enablelog", true, "Enable system wide logging, set to false for writing log to STDOUT only")
 
+	/* Statistics Configuration Flags */
+	statsMaxEntriesPerMap = flag.Int("stats_max_entries", 0, "Cap per-dimension statistics maps at N entries; when the cap is exceeded the entries with the lowest request counts are dropped first (0 = unbounded, matches upstream behavior; recommended value when enabled: 20000)")
+
 	/* Default Configuration Flags */
 	defaultInboundPort          = flag.Int("default_inbound_port", 443, "Default web server listening port")
 	defaultEnableInboundTraffic = flag.Bool("default_inbound_enabled", true, "If web server is enabled by default")
--- a/src/def.go
+++ b/src/def.go
@@ -85,7 +85,7 @@ var (
 	enableLog = flag.Bool("enablelog", true, "Enable system wide logging, set to false for writing log to STDOUT only")
 
 	/* Statistics Configuration Flags */
-	statsMaxEntriesPerMap = flag.Int("stats_max_entries", 0, "Cap per-dimension statistics maps at N entries; when the cap is exceeded the entries with the lowest request counts are dropped first (0 = unbounded, matches upstream behavior; recommended value when enabled: 20000)")
+	statsMaxEntriesPerMap = flag.Int("stats_max_entries", 0, "Soft-cap per-dimension statistics maps at N entries; when exceeded the entries with the lowest request counts are dropped first (0 = unbounded; recommended value when enabled: 20000)")
 
 	/* Default Configuration Flags */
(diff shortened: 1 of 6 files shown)


tobychui:
This is a lot of changes. I think I might need some time to review this properly. 

Rebasing to v3.3.4 for now.

ElmoViggiani:
I updated my branch with the requested Copilot changes. It compiles ok and passes the tests. I'm running it now in prod. My previous version with the fix was running for 10 days without any memory creep. It has been stable at 720 MiB on a busy site.

Sys.db is still 161MB as when I started, indicating the db entries are getting cleared as scheduled and boltdb is reusing the cleared space as expected.

Image 6-4-26 at 15 41

tobychui:
Thanks for the followup! Will be merging this and deploy to my dev branch this weekend 👍🏻

tobychui (approved):
LGTM

PR #1197, opened 2026-06-01
feat: per-route OAuth2 provider + auth exception rules for all auth m…

### LINKED ISSUE #1162: [ENHANCEMENTS] Path‑Level OAuth2 Exemptions
I would like to request an enhancement to the current OAuth2 enforcement mechanism. At the moment, enabling OAuth2 protection applies authentication requirements globally to all routes under a given domain. This creates operational issues for systems that rely on automated or internal API communication, since these endpoints cannot complete user‑interactive OAuth2 flows.

**Proposed Improvement**

Introduce support for path‑level exemptions within the OAuth2 configuration. This would allow administrators to enforce OAuth2 authentication for the main application while explicitly excluding specific routes (e.g., /api/*) from OAuth2 validation. Possible implementation approaches include:

- A list of regex‑based or prefix‑based path exclusions
- Per‑route OAuth2 toggles within routing rules
- A rule‑evaluation layer that determines whether OAuth2 should be applied before forwarding the request

**Rationale**

Modern architectures frequently separate frontend user interfaces (which require OAuth2 authentication) from backend services that communicate programmatically. Without the ability to define exceptions, enabling OAuth2 can unintentionally break internal API calls, service‑to‑service communication, mobile app requests, and webhook integrations.

Adding path‑level OAuth2 exemptions would significantly improve deployment flexibility and align Zoraxy with common reverse‑proxy and identity‑gateway patterns used in production environments.

Another important change is identifying the email that was authenticated and allowing only registered email addresses to have access.

Thank you for your continued work on the project.

BluLupo:
…ethods

- Add OAuth2ProviderConfig struct allowing each proxy route to use its own OAuth2/OIDC provider (different Authentik application, client ID and secret per virtual host)
- Apply auth exception rules (path prefix / CIDR bypass) to OAuth2, Forward Auth and ZorxAuth — previously only Basic Auth respected them
- Refactor HandleOAuth2Auth to merge per-route config over global settings with isolated cache keys (no cross-route pollution)
- Fix missing </div> in basicAuthEditor.html that hid the Add Exception button when Path Prefix mode was selected
- UI: show per-route OAuth2 config panel when OAuth2 auth method is selected; hide credentials section for non-Basic-Auth methods

Closes #1162

### DIFF
--- a/src/mod/auth/sso/oauth2/oauth2.go
+++ b/src/mod/auth/sso/oauth2/oauth2.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/jellydator/ttlcache/v3"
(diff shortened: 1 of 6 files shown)


tobychui:
Hi @kjagosz ,

As I remember you are the original author for this module, would you like to take a look at the changes before I merge this?

BluLupo:
The code would need to be reviewed, and some real-world testing would need to be carried out.

AnthonyMichaelTDM:
> The code would need to be reviewed, and some real-world testing would need to be carried out.

Does that mean you haven't tested these changes yourself yet? 

I'd like you to do that, and write some unit tests

BluLupo:
Hi, I've given up on it for now, but I was using it for work. The goal was to be able to authenticate certain routes with LDAP and LDAP group. What kind of tests would you like?

PR #1211, opened 2026-06-20
Fix Plugin Manager UI: iframe height and Developer Settings accordion

ElmoViggiani:
## Summary

Two small, web-asset-only fixes for the Plugin Manager UI.

### 1. Plugin UI iframe sized to viewport height
`resizeIframe()` in `src/web/components/plugincontext.html` sized the plugin iframe to the left sidebar's (`#mainmenu`) height, using the viewport-based height only as a fallback when the sidebar measured 0. On browser windows taller than the sidebar, the Zoraxy footer filled the leftover space, shrinking the usable plugin area to less than half the window height.

Now sizes to the larger of the sidebar height and the viewport height, so the iframe fills the window on tall browsers. It already re-runs on window resize and tab switch.

### 2. Developer Settings accordion never expanded
The "Developer Settings" accordion in `src/web/components/plugins.html` (`div.ui.accordion.advanceSettings`) was never initialized, so clicking it did nothing — and produced no console error, since an uninitialized Fomantic accordion is inert markup with no click handler bound. Every other component using this pattern calls `$(".advanceSettings").accordion();` (redirection, access, status, sso); `plugins.html` was the only one missing it.

Added the initialization alongside `initDeveloperSettings()`.

## Testing
Built and deployed to a live Zoraxy instance; both behaviors confirmed fixed in-browser (accordion expands/collapses; plugin iframe fills tall windows). Web UI is served from embedded assets (`//go:embed web/*`), so this requires a rebuild to take effect — no Go logic changed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/web/components/plugincontext.html
+++ b/src/web/components/plugincontext.html
@@ -48,7 +48,9 @@
         if (mainMenuHeight == 0){
             mainMenuHeight = window.innerHeight - 198; //Fallback to window height
         }
-        iframe.style.height = mainMenuHeight + 'px';
+        //Size to the taller of the sidebar and the viewport so the iframe
+        //fills the window on tall browsers instead of leaving a footer gap
+        iframe.style.height = Math.max(mainMenuHeight, window.innerHeight - 198) + 'px';
     }
 
     $(window).on("resize", function(){
--- a/src/web/components/plugincontext.html
+++ b/src/web/components/plugincontext.html
@@ -45,9 +45,6 @@
     function resizeIframe() {
         let iframe = document.getElementById('pluginContextLoader');
         let mainMenuHeight = document.getElementById('mainmenu').offsetHeight;
-        if (mainMenuHeight == 0){
-            mainMenuHeight = window.innerHeight - 198; //Fallback to window height
-        }
         //Size to the taller of the sidebar and the viewport so the iframe
         //fills the window on tall browsers instead of leaving a footer gap
         iframe.style.height = Math.max(mainMenuHeight, window.innerHeight - 198) + 'px';
--- a/src/web/components/plugins.html
+++ b/src/web/components/plugins.html
@@ -771,10 +771,8 @@ function uninstallPlugin(pluginId, pluginName, btn=undefined) {
   }
 
   initDeveloperSettings();
-  // Scope the accordion init to this page's Developer Settings only.
-  // Webmin components are loaded dynamically via ajax and several of them
-  // (redirection, status, access) reuse the .advanceSettings class, so a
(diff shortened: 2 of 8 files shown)


ElmoViggiani:
Plugin Manager iframe was too small on my 4k screen.
And the Plugin Manager -> Developer Settings accordion didn't expand.
Image 6-20-26 at 18 26

src/web/components/plugins.html
```diff
@@ -771,6 +771,7 @@ <h4 class="ui header">
   }
 
   initDeveloperSettings();
+  $(".advanceSettings").accordion();
```
tobychui: There are a few `.advanceSettings` class in the webmin panel due to the design that the contents are dynamically loaded via $ajax. You need to be more specific on which advance settings to toggle accordion to prevent duplicate calls to other page's elements here.

src/web/components/plugincontext.html
```diff
@@ -48,7 +48,9 @@
         if (mainMenuHeight == 0){
```
tobychui: Since your logic already use Math.max which covered the case where mainMenuHeight is 0 (i.e. in this case the window.inner height - 198px must be larger than 0), you should also remove this redundant if case.

tobychui (changes requested):
Hey, thanks for the PR!

A few places need some final touch-up, and this should be good to merge.

Also, it seems that you are using a low effort mode for submitting changes, next time please consider using a higher effort mode for proposing changes which properly remove functional duplication and avoid introducing other potential issues.

src/web/components/plugins.html
ElmoViggiani: Fixed in latest commit.
redirection.html, status.html, and access.html also use the unscoped $(".advanceSettings").accordion() and have the same latent issue.

src/web/components/plugincontext.html
ElmoViggiani: Fixed in latest commit.

ElmoViggiani:
Thanks for the tip. I will investigate my settings.

The accordion fix also applies to 3 other pages. Should I submit another PR for those? Or add them to this one?

redirection.html, status.html, and access.html still use the unscoped $(".advanceSettings").accordion() and have the same latent issue.

tobychui:
> The accordion fix also applies to 3 other pages. Should I submit another PR for those? Or add them to this one?

I think those need to be updated as well, you can do them in this PR if you wish, or I can fix them later.

ElmoViggiani:
Added the fix to the other 3 pages.

tobychui (approved):
LGTM

PR #1210, opened 2026-06-20
feat(forward-auth): add Ignored Paths 

CrazyWolf13:
## What this does

Fixes the Authentik single-application forward-auth redirect loop (#895) by adding a generic Ignored Paths option to the (global) Forward Auth settings, a comma-separated list of request path prefixes that bypass the forward-auth check entirely.

## Setup (Authentik single-application mode)
1. Forward Auth -> Advanced Options -> Ignored Paths = `/outpost.goauthentik.io`
2. A virtual directory `/outpost.goauthentik.io` -> `<authentik-host>:9000/outpost.goauthentik.io` (Require TLS off) on each protected host, routes the OAuth callback to the outpost.

Both the Address and Ignored Paths fields carry inline notes guiding this, so it should be self-explanatory and no wiki page should be needed.

> [!WARNING]
> Paths in **Ignored Paths** get **no authentication whatsoever** — any request whose path starts with one of those prefixes bypasses Forward Auth entirely. Only list auth-callback paths (e.g. `/outpost.goauthentik.io`) or paths you intentionally want public. (This warning is also shown in the UI next to the field.)

grafik

## Implementation (AI-Assisted:)

- **`forward` package:** new `IgnoredPaths` option (load / GET / POST / DELETE / log) + `IsIgnoredPath()` with
hardened, normalized matching (decoded + `path.Clean` + boundary check) so `..`/`%2e%2e` traversal and prefix-boundary
tricks can't bypass auth on unintended paths. Empty entries are skipped so a stray comma can't disable auth
site-wide.
- **`dynamicproxy/authProviders.go`:** `handleForwardAuth` returns early for ignored paths — modeled on the existing
`ZorxAuthExceptionRules` pattern. **~6 lines, no `Server.go`/routing changes**, so it stays clear of the dynamic-proxy
work on `v3.3.4`.
- **`web/components/sso.html`:** Ignored Paths field under Advanced Options, with the inline warning + Authentik hint.
- Unit tests for the matcher (boundary, traversal, multi-prefix).

## Testing
test seem to pass, builds fine and I tested this build succesfully for multiple hours and will keep it live in my homelab, though I'd appreciate any testing!

## Open question
Currently the `/outpost.goauthentik.io` vdir needs to be added manually per host. 
I'd happily add a tickbox or similar to automatically handle this, but I first wanted to hear your opinion on this and kept the PR to a basic level.

> [!NOTE]
[...]

PR #1214, opened 2026-06-23
feat(forward-auth): bulk apply/remove vdir for forward auth hosts

CrazyWolf13:
Continuation of #1210

Adds an easy way to create the virtual directory that Authentik single-application Forward Auth needs, instead of adding it by hand on every host.

From Forward Auth -> Advanced Options, two buttons manage the callback vdir across all hosts by their Forward Auth state:

- Add: create it on every host that uses Forward Auth
- Remove: delete it from every host that doesn't (cleanup)

Conflict-safe, an existing directory is only skipped/removed when its target matches exactly; one at the same path with different settings is reported as a conflict and left untouched. A per-host summary (created / removed / skipped / conflicts) is shown after each run.

Adding
removing

Removing
adding

> [!NOTE]
> I'm not a professional dev and this was written largely with AI assistance, but was well vetted and tested. I'd be very happy if you can take a look and give it a thorough review!

### DIFF
--- a/src/api.go
+++ b/src/api.go
@@ -64,6 +64,7 @@ func RegisterHTTPProxyAPIs(authRouter *auth.RouterDef) {
 	authRouter.HandleFunc("/api/proxy/vdir/add", ReverseProxyAddVdir)
 	authRouter.HandleFunc("/api/proxy/vdir/del", ReverseProxyDeleteVdir)
 	authRouter.HandleFunc("/api/proxy/vdir/edit", ReverseProxyEditVdir)
+	authRouter.HandleFunc("/api/proxy/vdir/bulkForwardAuth", ReverseProxyBulkApplyVdirByForwardAuth)
 	/* Reverse proxy user-defined header */
 	authRouter.HandleFunc("/api/proxy/header/list", HandleCustomHeaderList)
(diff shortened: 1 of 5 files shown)


tobychui:
I think the scope of which this PR touches is a bit large, let me first release rc-1 and do some in depth testing in rc-2 for these changes.

CrazyWolf13:
sure, yeah it got sadly a bit bigger than I imagined/wanted it to get.
If you find a better way to implement the same functionality with less code, let me know.

tobychui (approved):
LGTM, but there is one minor concern over the `dynamicproxy.BulkVdirRemove` function. 

It seems it didn't handle the case where the vdir is created with existing alternative upstream (i.e. there already exists a vdir that point to somewhere else), then when it is deleted, the vdir that is **not** created by previous bulk create will get removed too since the remove didn't check if the vdir being remove is exactly the vdir that pointing to the SSO endpoint configured by user (I am not expert in forward auth, maybe not the correct term here)

But I guess people who use forward auth should be smart enough to figure these out, so I will leave it as is right now, maybe worth a quick patch later to optimize user experience on this design.

CrazyWolf13:
thanks for the review and good spot, didn't think about this!

I created a small follow-up: https://github.com/tobychui/zoraxy/pull/1237

PR #1218, opened 2026-06-27
tlscert: use name-based fallback cert instead of file renaming

zen8841:
refactor: replace file-renaming with name-based fallback cert
Instead of renaming certificate files to default.pem/key when
setting a fallback cert, store the cert name in Manager.FallbackCert
and persist it in fallback.json. This eliminates the root cause
of ACME DNS credential mismatches — since files are never moved,
database keys always stay in sync with filenames.

Changes:
- Add FallbackCert field to Manager struct
- Rewrite SetCertAsDefault to only store the name, no file renaming
- Update DefaultCertExists/Sep and TLS fallback to use FallbackCert
- Add v334 migration for legacy default.* files
- Read fallback.json on startup to restore the fallback cert name

Fix the issue raised by #1181 #1124

I've already tested it, but just to be safe, @tobychui  , could you test it ?

### DIFF
--- a/src/mod/tlscert/handler.go
+++ b/src/mod/tlscert/handler.go
@@ -43,24 +43,24 @@ func (m *Manager) HandleCertDownload(w http.ResponseWriter, r *http.Request) {
 		utils.SendErrorResponse(w, "invalid certname given")
 		return
 	}
-	certname = filepath.Base(certname) //prevent path escape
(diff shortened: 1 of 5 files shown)


tobychui (approved):

tobychui:
Looks good to me, will be merging this first and I will test it together with my work in progress TLS module update ✨

PR #1221, opened 2026-06-30
fix 1217: frontend js forgot to call parameter

### LINKED ISSUE #1217: [BUG] Preferred CA is ignored when requesting a cert during new proxy host creation
**Describe the bug**
When requesting a certificate through the confirmation prompt after creating a new proxy host ("Request new SSL Cert for this subdomain?" prompt) the Preferred CA set on the ACME Client page is silently ignored, and the request is sent to the backend with an empty `ca` parameter, so the backend falls back to the default Let's Encrypt. The UI reports success while the server log and new certificate shows Let's Encrypt.

**To Reproduce**
1. Go to the ACME Client page and set Preferred Certificate Authority to ZeroSSL and save.
2. Go to Create Proxy Rule page and create a new HTTP proxy host.
3. After host is created, a dialog ("Request new SSL Cert for this subdomain?") will pop up, select the check mark to request the certificate.
4. Watch server logs, it will print `CA not set. Using default` and issues from Let's Encrypt instead of ZeroSSL. View the certificate after it is created and it is from Let's Encrypt, not ZeroSSL.

**Expected behavior**
The certificate request triggered during new-host creation should use the Preferred Certificate Authority selected. With ZeroSSL selected, the backend should issue from ZeroSSL on confirmation of the dialog

**Screenshots**
Image
Image

```log
[internal] [system:info] Updating prefered ACME CA to ZeroSSL
[ACME] [system:info] CA not set. Using default
[ACME] [system:info] Obtaining certificate for: example.com
[ACME] [system:info] Using https://acme-v02.api.letsencrypt.org/directory for CA Directory URL
```

**Browser (if it is a bug appears on the UI section of the system):**
 - OS: Ubuntu 24.04
 - Browser Firefox
 - Version 152.0.1

**Host Environment (please complete following information, DO NOT REMOVE ANY FIELD(S)):**
-  Arch: x86_64
 - Device: Proxmox LXC container
 - OS: Debian
 - Version 13.5 Trixie
 - Are you using Docker? no
 - Docker Version: N/A
 - Zoraxy Version: 3.3.3

**Additional context**
The issue stems from `src/web/components/rules.html` calling `obtainCertificate` incorrectly. 

https://github.com/tobychui/zoraxy/blob/ddf62ad3ae709749f39ad9736f0c29666b78516b/src/web/components/rules.html#L433

This can be fixed by adding in the `dns` parameter, like in `cert.html`.

https://github.com/tobychui/zoraxy/blob/ddf62ad3ae709749f39ad9736f0c29666b78516b/src/web/components/cert.html#L187

Adding in `false` as the DNS parameter gets ZeroSSL to work

```log
[ACME] [system:info] Obtaining certificate for: example.com
[ACME] [system:info] Using https://acme.zerossl.com/v2/DV90 for CA Directory URL
```

zen8841:
fix #1217 
frontend js forgot to call parameter

The #rules page only allows the use of the HTTP 01 challenge. I don't have the environment to test this, but it calls the ACME function, which is consistently used, so there shouldn't be any issues.

### DIFF
--- a/src/web/components/rules.html
+++ b/src/web/components/rules.html
@@ -438,7 +438,7 @@
                                 console.log("Trying to get a new certificate via ACME");
 
                                 //Request ACME for certificate, see cert.html component
-                                obtainCertificate(rootname, defaultCA.trim(), function(){
+                                obtainCertificate(rootname, false, defaultCA.trim(), function(){
                                      // Renew the parent certificate list
                                     initManagedDomainCertificateList();
                                 });


tobychui (approved):
I think the renew button only support HTTP challenge is what the previous ACME maintainer preferred design. I will test this later after rc-1 release. Thanks!

PR #1227, opened 2026-07-05
Docker hardening

PassiveLemon:
Remove Alpine apk binary from image

Fix some CVEs:
x/net v0.54.0 -> v0.55.0

Updated Docker to v29:
docker/docker -> moby/moby/client v0.3.0
Removed docker/docker/api
Fixed docker.go response handling

Updating gorilla/csrf to v1.7.3 would fix another CVE, but that breaks account creation for some reason.

Additionally but irrelevant, Zerotier does not work on v3.3.4-rc1. I'm able to see my Network Controller ID, but using create network errors: `Network error. status code: 404`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,15 +39,6 @@ src/sys.uuid
 src/zoraxy
 src/log/
 
-# dev-tags
-/Dockerfile
-/Entrypoint.sh
-
-# docker testing stuff
-docker/test/
-docker/container-builder.sh
-docker/docker-compose.yaml
-
 # plugins
 example/plugins/ztnc/ztnc.db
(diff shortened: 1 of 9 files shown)


tobychui:
I guess it was caused by ZeroTier implementation update that works for newer version of ZeroTier instead of the old version we have in docker? Maybe we could remove ZeroTier from the docker from next release onward, I might take a look at the possible to integrate Tailscale or a custom implementation of GAN like features.

PassiveLemon:
I'll mess around with the ZeroTier version and see if newer versions work. I wouldn't mind removing it all together because it's kind of annoying but I guess that would depend on how many people are even using the ztnc plugin

Issue #1088, opened 2026-03-23
[BUG] Error 521, after Upgrade to 3.3.2


worldwidehigh:
**Describe the bug**
After Upgrade to 3.3.2. I became Error 521. When I move back its ok. Its only one service ([Omada](https://github.com/mbentley/docker-omada-controller)).

**To Reproduce**
1. Upgrade to 3.3.2.
2. Error 521
3. Move back to 3.3.1
4. Its all ok

**Expected behavior**
The Service is online

**Screenshots**
<img width="987" height="624" alt="Image" src="https://github.com/user-attachments/assets/92720dd8-6a08-49b6-a44b-f87224977769" />

**Host Environment (please complete following information, DO NOT REMOVE ANY FIELD(S)):**
-  Arch: amd64
 - Device: VPS
 - OS: Ubuntu
 - Version: 24.04
 - Are you using Docker? yes
 - Docker Version 29.3.0
 - Zoraxy Version 3.3.2.

cft3ck:
Same error, same versions. Upgrading to v3.3.2 broke connections to proxy targets that require TLS (such as AdguardHome for the DoH feature). Selecting "Skip Verification" worked momentarily, but eventually stop working completely.  Reverting back to v3.3.0 fixed the issue. 

<img width="1051" height="632" alt="Image" src="https://github.com/user-attachments/assets/8bf55330-dda5-4043-8382-61932350fba3" />

Zoraxy logs were just 521 errors and also flooded the AdguardHome logs with TLS handshake errors. 

```
2026-03-23T08:29:04.010810832Z 2026/03/23 08:29:04.010705 [error] http: TLS handshake error from 172.18.0.6:57234: remote error: tls: bad certificate server=https
```
Note: the certs used by Zoraxy and AdguardHome are the same (generated by the ACME Auto renew tool). 

- Arch: amd64
- Device: Mini PC Intel(R) Celeron(R) N5095A
- OS: Debian
- Version: 13
- Are you using Docker? yes
- Docker Version 29.2.1, build a5c7197
- Zoraxy Version 3.3.2

tobychui:
@cft3ck I am not sure if your issue is the same as @worldwidehigh one, but a lot of information are missing here and just one line of log didn't help much with the debugging. 

Please post the full log section where the request is being made, preferably with curl and wireshark capture results.

cft3ck:
[...]

PR #1228, opened 2026-07-08
Fix: restore per-request TLS SNI for IP-addressed HTTPS upstreams

ElmoViggiani:
I have multiple upstreams on wireguard, reachable via IP address, 10.10.1.x.
There's a regression in v3.3.4 that doesn't send SNI to upstreams that are IP's.
I used Claude to track it down and fix it.

## Problem

Since #1088 (shipped in v3.3.4-rc1), the upstream TLS Server Name Indication
(SNI) is set once at transport-creation time instead of per request. The fixed
`ServerName` is `target.Hostname()`.

When an upstream is addressed by **IP** — e.g. a backend reachable only over a
tunnel that still serves a certificate for the proxied domain — that
`ServerName` is an IP literal. Go's `crypto/tls` omits IP-literal server names
from the ClientHello entirely (per RFC 6066), so the backend receives **no SNI**
at all. An SNI-routed server with no default virtual host then rejects the
handshake, and Zoraxy returns an instant **521** for every such endpoint even
though the upstream is perfectly healthy.

This is a regression: before #1088 the SNI was derived from the request host on
each request, so IP upstreams worked.

## Fix

Restore the pre-#1088 behavior **for IP upstreams only**:

- **Hostname upstreams** (the common case) are unchanged — they keep the fixed
  shared `ServerName` set at construction time, and therefore keep the
  connection-pooling improvement from #1088.
- **IP upstreams** leave the shared `ServerName` empty and derive the SNI from
  the (possibly rewritten) request host on each request. The per-request path
  clones the transport only when needed and skips the override when the derived
  name is empty or itself an IP (which Go would drop anyway).

## Tests

Adds `dpcore_sni_test.go`:

- `TestInitTimeSNI` — a hostname upstream keeps a fixed shared SNI, an IP
  upstream leaves it empty, and an HTTP upstream has no TLS config.
- `TestIPUpstreamUsesRequestHostSNI` — end-to-end regression test: an
  IP-addressed HTTPS upstream (a `httptest` TLS server on `127.0.0.1`) receives
  the proxied request host as its SNI via `GetConfigForClient`.

Both pass with `go test ./mod/dynamicproxy/dpcore/`.

## Scope

Two files, `src/mod/dynamicproxy/dpcore/dpcore.go` (+the new test). No config or
API surface changes; hostname-upstream behavior is byte-for-byte identical.


tobychui:
Hi @ElmoViggiani 

Thanks for the PR!

Generally I don't recommend using AI tool to make large modification to dpcore, as changes in dpcore has to be carefully reviewed before merging. Since this change is kinda small, I have review it manually and it should be fine for merging without introduce another breaking changes. 

Next time if any changes to dpcore are needed, please consider opening a issue explaining in details why it is needed and how the potential fix can help with the issue, so maintainers can review it beforehand. Thank you!

tobychui (approved):
LGTM

ElmoViggiani:
Understood. Thank you.

PR #1234, opened 2026-07-17
Feature (acme): custom acme preferred ca

### LINKED ISSUE #383: [ENHANCEMENTS] Possibility to Use Custom Preferred Certificate Authority
I am using a custom PKI with self-signed certificates, so it would be beneficial to add the option to use a custom ACME provider for certificate generation and renewal. Currently, only the ACME Tool allows the use of custom ACME servers, which suggests that the programmatic logic is already in place.

### LINKED ISSUE #1084: [ENHANCEMENTS]Add Option for Custom Preferred Certificate Authority for Auto‑Renewal
Within Zoraxy, there is already functionality to use the ACME tool to generate custom SSL certificates for subdomains. I’d like to request a feature enhancement on the TLS/SSL Certificates page: the addition of a “Custom Preferred Certificate Authority” option (e.g., Smallstep‑CA) that can be used for automated certificate renewals. This would remove the need to manually run the ACME tool each time a certificate expires.

Currently, only Let’s Encrypt and ZeroSSL are available for automated renewal. Since Zoraxy already supports issuing certificates from a custom CA through the ACME tool, extending this capability to the certificate‑renewal workflow would significantly streamline management.

Image

### LINKED ISSUE #1199: [ACME] Feature Request: Allow custom ACME server URL in Preferred CA dropdown
**Describe the bug**
This is a feature request related to the ACME flow.

The "Preferred Certificate Authority" dropdown only lists hardcoded public CAs 
(Let's Encrypt, ZeroSSL, Buypass, Google). There is no way to set a custom 
ACME server as the default CA for automatic certificate generation on new proxy rules.

Custom ACME servers already work correctly in "Generate New Certificate" → 
"Custom ACME Server", and the CA URL is properly saved in the per-certificate 
`.json` sidecar file for auto-renewal. The gap is only in the Preferred CA 
global setting.

**Request Details**
- ACME mode: HTTP challenge (auto-approved on internal network)
- CA used: Self-hosted custom ACME server (RFC 8555) bridged to internal PKI
- DNS provider: N/A

**Error Message**
N/A — this is a feature request, not a bug. The current behavior is that the 
Preferred CA dropdown only shows public CAs hardcoded in `src/mod/acme/ca.json` 
(embedded in the binary). There is no way to extend this list.

**To Reproduce**
1. Go to Certificates → CA and Auto Renew (ACME)
2. Open the "Preferred CA" dropdown
3. Observe that only Let's Encrypt, ZeroSSL are available
4. There is no "Custom" option or way to add a custom ACME directory URL

**Expected behavior**
Add a "Custom" option in the Preferred CA dropdown (similar to what already 
exists in "Generate New Certificate"), so users can set a custom ACME directory 
URL as the global default.

Alternatively: support a user-provided `ca.json` file in the config directory 
that merges with or overrides the embedded one.

**Use case**
Internal infrastructure where domains cannot be validated by public CAs 
(e.g. `*.company.intra`). A self-hosted ACME bridge connected to an internal 
PKI (Infisical, Step-CA, CFSSL…) can issue certificates for these domains — 
but it cannot be set as the preferred CA for new proxy rules.

**Host Environment**
- Arch: amd64
- Device: VM
- OS: Debian 12
- Docker: yes
- Docker Version: 27.x
- Zoraxy Version: latest (zoraxydocker/zoraxy:latest)

jacosta57:
The Preferred CA setting only supported the CAs hardcoded in ca.json. The Generate New Certificate tool already supports custom ACME directory URLs, but there was no way to make a custom server the default CA.

This PR adds the ability to add a custom CA URL, as well as set EAB credentials for both custom and for ZeroSSL certs.

Closes #1199, closes #1084, closes #383

## Changes
The GET response of `/api/acme/autoRenew/ca` changed from a bare string `"Let's Encrypt"` to an object. The only consumer, `acme.html`, is updated in this PR. 

`IsSupportedCA` was previously being called but not used, saving an unsupported CA name is now rejected at save

Switching from custom back to a named CA keeps the stored URL, so switching back restores the URL, consumers only read the URL when CA is `custom`

## Screenshots
image
image

### DIFF
--- a/src/acme.go
+++ b/src/acme.go
@@ -6,9 +6,11 @@ import (
 	"io"
 	"math/rand"
 	"net/http"
+	"net/url"
 	"os"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	"imuslab.com/zoraxy/mod/acme"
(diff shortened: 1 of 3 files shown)


tobychui:
Thanks for the PR! 

Changes looks legit, will wait for @zen8841 review before merge.

zen8841 (approved):
LGTM
Although I still have a few ideas for improvement, let's merge it for now

BillyOne971:
Merci 😉

PR #1246, opened 2026-07-29
chore: update to lego v5

### LINKED ISSUE #1245: [BUG] Infomaniak DNS Challenge fails: "cannot unmarshal number into Go struct field APIResponse[string].data of type string"
### Describe the bug
When using the ACME Client's DNS Challenge with Infomaniak as the DNS Provider, certificate issuance fails after the DNS TXT record has already been successfully created via the Infomaniak API. The error occurs while parsing the API response.

### To Reproduce
1. Go to TLS/SSL Certificates → Open ACME Tool
2. Select "Use a DNS Challenge"
3. Select "Infomaniak" as DNS Provider
4. Fill in:
   - APIEndpoint: `https://api.infomaniak.com`
   - AccessToken: (valid Infomaniak API token with Domain scope)
   - PollingInterval: 2
   - PropagationTimeout: 600
5. Click "Get Certificate"

### Error message

Error: one or more domains had a problem: [patking73.ch] [patking73.ch]
acme: error presenting token: infomaniak: error when calling api to create DNS record:
unable to unmarshal response: [status code: 201] body:
{"result":"success","data":36351066}
error: json: cannot unmarshal number into Go struct field APIResponse[string].data of type string

### Analysis
The Infomaniak API responds with HTTP 201 and a body confirming success:
```json
{"result":"success","data":36351066}
```
The `data` field here is a **numeric ID** (the ID of the newly created DNS record), not a string. The underlying `APIResponse` struct (likely in the embedded `lego` Infomaniak DNS provider) appears to expect `data` to always be of type `string`, causing the JSON unmarshal to fail even though the DNS record was created successfully.

This means the DNS-01 challenge TXT record *is* being created correctly, but Zoraxy/lego aborts right after due to this parsing error, so the certificate is never issued.

### Suggested fix
Change the `data` field type in the `APIResponse` struct to `json.RawMessage` or `interface{}` (or `any`) instead of `string`, since the Infomaniak API appears to return different types depending on the endpoint/operation (e.g. string for some responses, numeric ID for record creation).

### Related discussion
This appears to be the same underlying class of issue reported in #475 ("DNS Challenge and Infomaniak not working"), which was left open due to lack of maintainer familiarity with the Infomaniak provider and no clear repro/root cause at the time. This report provides that missing detail.

### Proof the Infomaniak API itself works fine
Infomaniak maintains their own official certbot plugin, `certbot-dns-infomaniak` (https://github.com/Infomaniak/certbot-dns-infomaniak), which successfully completes DNS-01 challenges against the same API. This confirms the API and token setup are correct, and that the issue is isolated to response parsing in Zoraxy's (or the embedded lego's) Infomaniak DNS provider implementation, not the Infomaniak API itself or user configuration.

### Host Environment
- Zoraxy version: v3.3.3
- Deployment: LXC container (Proxmox)
- DNS Provider: Infomaniak

### Expected behavior
Certificate should be issued successfully via DNS-01 challenge against Infomaniak.

ldez:
Hello, I'm the lego maintainer.

This PR is only a technical update to lego v5, I ran all the existing tests.
But I didn't run zoraxy to check the impact of the migration.

I do this PR only because one zoraxy user is annoying me with low value AI generated stuffs.

Fixes #1245

Note: I also fixed an unrelated test.


tobychui:
Hi @ldez 

Thanks for the help!

Yeah these kind of AI generated PR happens quite often recently, I will let our ACME maintainer take a look at it real quick.

Hi @zen8841 

Would you mind take a look at this and see if this works out? Many thanks!

zen8841:
Thanks for the PR.
Adn sorry about the trouble from those AI-generated issues, https://github.com/go-acme/lego/issues/3216 https://github.com/go-acme/lego/issues/3217

Also, I noticed that in lego v4, the pre-check only used the Authoritative NS:
```
func newPreCheck() preCheck {
    return preCheck{
        requireAuthoritativeNssPropagation: true,
    }
}
```
I tested this PR, and using the Recursive NS for the DNS propagation check takes too long — it can exceed the timeout. I'm planning to change it to DisableRecursiveNSsPropagationRequirement() and use the Authoritative NS for the pre-check instead.

ldez:
I made a mistake, you should replace `DisableAuthoritativeNssPropagationRequirement` by `DisableRecursiveNSsPropagationRequirement`

PR #1256, opened 2026-08-01
feat: WebAuthn and TOTP 2FA support for admin interface

mvasl:
Added a 2-factor authentication feature via TOTP-based verification codes and passkeys support via WebAuthn for the admin panel.

WebUI part is mostly generated via Qwen3.6 since I'm not familiar with jquery, please don't kick me too hard if it is bad in some way.

### DIFF
--- a/src/mod/auth/auth.go
+++ b/src/mod/auth/auth.go
@@ -6,16 +6,20 @@ package auth
 */
 
 import (
+	"bytes"
 	"crypto/rand"
 	"crypto/sha512"
+	"encoding/base64"
+	"encoding/hex"
 	"errors"
+	"fmt"
+	"image/png"
 	"net/http"
 	"net/mail"
 	"strings"
 
-	"encoding/hex"
-
 	"github.com/gorilla/sessions"
+	"github.com/pquerna/otp/totp"
 	db "imuslab.com/zoraxy/mod/database"
 	"imuslab.com/zoraxy/mod/info/logger"
 	"imuslab.com/zoraxy/mod/utils"
@@ -122,10 +126,23 @@ func (a *AuthAgent) HandleLogin(w http.ResponseWriter, r *http.Request) {
 	//The database contain this user information. Check its password if it is correct
 	if passwordCorrect {
 		//Password correct
-		// Set user as authenticated
-		a.LoginUserByRequest(w, r, username, rememberme)
+		if a.IsTOTPEnabled(username) {
+			// Set session data but keep authenticated=false while waiting for TOTP
+			session := a.PrepareUserSession(w, r, username, rememberme)
+			session.Values["totp_partial"] = true
+			session.Values["authenticated"] = false
+			session.Save(r, w)
 
-		//Print the login message to console
+			utils.SendJSONResponse(w, `{"totp_required":true,"username":"`+username+`"}`)
+			a.Logger.PrintAndLog("auth", username+" password verified, waiting for TOTP code.", nil)
+			return
+		}
+
+		// No TOTP - fully authenticate
+		session := a.PrepareUserSession(w, r, username, rememberme)
+		session.Values["authenticated"] = true
+		session.Save(r, w)
+
 		a.Logger.PrintAndLog("auth", username+" logged in.", nil)
 		utils.SendOK(w)
 	} else {
@@ -160,11 +177,10 @@ func (a *AuthAgent) ValidateUsernameAndPasswordWithReason(username string, passw
 	}
 }
 
-// Login the user by creating a valid session for this user
-func (a *AuthAgent) LoginUserByRequest(w http.ResponseWriter, r *http.Request, username string, rememberme bool) {
+// Prepare valid session with user parameters. Caller is expected to commit the data with session.Save() call.
+func (a *AuthAgent) PrepareUserSession(w http.ResponseWriter, r *http.Request, username string, rememberme bool) *sessions.Session {
 	session, _ := a.SessionStore.Get(r, a.SessionName)
 
-	session.Values["authenticated"] = true
 	session.Values["username"] = username
 	session.Values["rememberMe"] = rememberme
 
@@ -180,7 +196,7 @@ func (a *AuthAgent) LoginUserByRequest(w http.ResponseWriter, r *http.Request, u
 			Path:   "/",
 		}
 	}
-	session.Save(r, w)
+	return session
 }
 
 // Handle logout, reply OK after logged out. WILL NOT DO REDIRECTION
@@ -210,6 +226,7 @@ func (a *AuthAgent) Logout(w http.ResponseWriter, r *http.Request) error {
 	}
 	session.Values["authenticated"] = false
 	session.Values["username"] = nil
+	session.Values["totp_partial"] = false
 	session.Options.MaxAge = -1
 	return session.Save(r, w)
 }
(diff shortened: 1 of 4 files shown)


tobychui:
Hi @mvasl 

AI generated code is fine but I don't think 2FA shall be supported in the admin interface in the first place. 
My point is:

1. The admin interface shall not be setup in a way that it is accessible from the internet. It should only be accessible within LAN or within a trusted network (access via private VPN, tailscale or zerotier network for example)
2. For multi-user use case, you should add Zoraxy auth as a proper authentication provider (e.g. point `zoraxy.example.com` to `127.0.0.1:8000` and add Zoraxy auth TOTP 2FA to it, optionally running Zoraxy in "no auth" mode so you dont login twice)

So for case 1, you dont need 2FA and for case 2, Zoraxy auth have TOTP 2FA build-in already. So this feature seems redundant. .

mvasl:
Well, case 1 is not always true, since many people (including me) prefer to organize their systems in zero-trust manner rather than building guarded perimeters, 2FA starts to make sense if you look at it like that.

mvasl:
Also, if you think about it, case 2 also makes no sense without privilege separation.

Every one allowed to login is in full control and can edit Zoraxy Auth users, overtake the whole system, it's no better that sharing single user+password pair, but thats offtopic in this PR.

mvasl:
Added passkeys support with discoverable credentials via WebAuthn as well.

tobychui:
Sounds cool, let me know when you are done changing stuffs (or convert it to Draft for now if you still plan to make changes in the following week). After finish, please provide a few screenshots on the new login UI and setup page (webmin panel). This will greatly help with the workflow when I am integrating the v3.3.5-rc1 new features next month.

mvasl:
> Sounds cool, let me know when you are done changing stuffs (or convert it to Draft for now if you still plan to make changes in the following week).

The only change i plan is to add rate-limiting to the webauthn "begin" endpoint. It is not strictly needed, more like a safeguard.

UI screenshots:

mvasl:
@tobychui I think i'm done here. Can you please take a look?

tobychui:
This is a huge change, maybe next weekend.

Btw, what qwen parameter size you are using here? I think you need at least qwen3.6 35B to generate code with decent quality, so if it is generated using a model smaller than that I guess I might need to ask Claude for a advance review.

mvasl:
It was Qwen3.6 MoE model with 35B parameters, yes, but only the frontend is generated. I asked both qwen and gemma4 26B to review this as well, but findings were, lets just say hit or miss.

tobychui:
Ok no problem 👍🏻 
Also, maybe you could try to cleanup the PR a bit, I just have a quick look at the changed files, it seems you have accidentally commited a agent config file or something like that. Will have a detail look next weekend.

mvasl:
Yeah, looks like I accidentally committed some of my Zed debugger dotfiles, dropped it and squashed the commits.

tobychui:
It seems like you are still making small changes here and there. If you are still testing this on your end, please turn this PR into a draft and finish editing before setting it to "Ready for review".

mvasl:
It was just a small fine-tune of rate limiter

tobychui:
Cool then, I am rebasing this to v3.3.5 for the rc-1 release

mvasl:
> I am rebasing this to v3.3.5 for the rc-1 release

Looks like something gone wrong, there are some unrelated commits now

tobychui:
@mvasl That is normal, you also need to rebase this to v3.3.5 branch on your side so it can merge.

src/mod/auth/auth.go
```diff
@@ -344,142 +599,1147 @@ func (a *AuthAgent) CheckAuth(r *http.Request) bool {
...
+		//This user do not exists.
+		return errors.New("this user does not exists")
+	}
+
+	//OK! Remove the user from the database
+	a.Database.Delete("auth", "passhash/"+username)
+	a.Database.Delete("auth", "email/"+username)
+	a.Database.Delete("auth", "totp/secret/"+username)
+	a.Database.Delete("auth", "totp/enabled/"+username)
+	a.Database.Delete("auth", fmt.Sprintf("webauthn/creds/%s", username))
```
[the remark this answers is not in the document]
mvasl: I guess this is not something i should touch, this is a broken abstraction from out of scope of this PR.

src/mod/auth/auth.go
```diff
@@ -344,142 +599,1147 @@ func (a *AuthAgent) CheckAuth(r *http.Request) bool {
...
+		return
+	}
+
+	wa, err := newWebAuthnFromRequest(r)
+	if err != nil {
+		utils.SendErrorResponse(w, "webauthn init failed: "+err.Error())
+		return
+	}
+
+	user := &webAuthnUser{Username: username}
```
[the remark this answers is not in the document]
mvasl: Nope, it's fine, no panics as proven by the screenshots above

src/mod/auth/auth.go
```diff
@@ -344,142 +599,1147 @@ func (a *AuthAgent) CheckAuth(r *http.Request) bool {
...
+		a.Database.Delete("auth", fmt.Sprintf("webauthn/pending_reg/%s", token))
+		utils.SendErrorResponse(w, "registration session expired")
+		return
+	}
+
+	username, ok := pendingData["username"].(string)
+	if !ok {
+		utils.SendErrorResponse(w, "invalid registration session")
+		return
+	}
```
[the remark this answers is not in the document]
mvasl: The fact that zoraxy admin panel only allows to create one user makes this irrelevant

src/mod/auth/auth.go
```diff
@@ -344,142 +599,1147 @@ func (a *AuthAgent) CheckAuth(r *http.Request) bool {
...
+	sessionJSON, err := serializeSessionData(sessionData)
+	if err != nil {
+		a.Logger.PrintAndLog("auth", "failed to serialize webauthn session data", err)
+		utils.SendErrorResponse(w, "failed to prepare auth: "+err.Error())
+		return
+	}
+	a.Database.Write("auth", fmt.Sprintf("webauthn/pending_auth/%s", token), map[string]any{
+		"sessionData": sessionJSON,
+		"expiry":      time.Now().Add(5 * time.Minute).Format(time.RFC3339),
+	})
```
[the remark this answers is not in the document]
mvasl: Yeah, I guess I can move those to some in-memory cache with TTL on records.

@tobychui is there any other place in zoraxy that uses similar pattern? Would it be too much to hope that cache with TTL-based key expiration is already implemented somewhere in the project?

src/web/components/utils.html
```diff
@@ -636,4 +726,350 @@ <h3><i class="heart red icon"></i> Financial Supporters</h3>
...
+                    user: {
+                        id: base64UrlToBufferWebauthn(options.user.id),
+                        name: options.user.name,
+                        displayName: options.user.displayName,
+                    },
+                    challenge: base64UrlToBufferWebauthn(options.challenge),
+                    pubKeyCredParams: options.pubKeyCredParams,
+                    timeout: options.timeout,
+                    attestation: options.attestation,
+                };
```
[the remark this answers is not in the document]
mvasl: Seems valid, will mirror how it is done in zoraxy auth

src/web/components/utils.html
```diff
@@ -636,4 +726,350 @@ <h3><i class="heart red icon"></i> Financial Supporters</h3>
...
+                    .create({ publicKey })
+                    .then(function(credential) {
+                        // Step 2: Send credential to server
+                        const credName = prompt("Enter a name for this passkey (e.g., 'iPhone 15', 'Windows Hello'):",
+                            "Passkey " + (data.credentialCount || 1));
+                        if (!credName) {
+                          $("#passkeyRegMsg").text("Registration declined.").show();
+                          $("#registerPasskeyBtn").prop("disabled", false).html('<i class="ui plus circle icon"></i> Register New Passkey');
+                          return
+                        }
```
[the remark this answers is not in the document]
mvasl: Will just ask for name before registration. Will have to drop passkeys counter, but its ok, i guess.

src/mod/auth/auth.go
```diff
@@ -344,142 +599,1147 @@ func (a *AuthAgent) CheckAuth(r *http.Request) bool {
...
+	enabled := a.IsTOTPEnabled(username)
+	status := "disabled"
+	if enabled {
+		status = "enabled"
+	}
+	utils.SendJSONResponse(w, `{"status":"`+status+`"}`)
+}
+
+// HandleTOTPGenerate generates a new TOTP secret and QR code for the logged-in user
+func (a *AuthAgent) HandleTOTPGenerate(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
mvasl: Fixed

src/statuspage.go
```diff
@@ -25,63 +25,230 @@ import (
...
 /* Bandwidth (Today) tracker */
 
-// dashboardBandwidthTracker keeps the NIC accumulated counter values sampled
-// at startup / midnight so that "bandwidth today" can be derived from the
-// current accumulated counters without a dedicated background routine.
+const (
+	//Traffic accumulation sample interval
+	DASHBOARD_BANDWIDTH_SAMPLE_INTERVAL = 5 * time.Second
+	//Auto save interval
+	DASHBOARD_BANDWIDTH_SAVE_INTERVAL = 60 * time.Second
```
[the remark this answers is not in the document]
mvasl: Unrelated

mvasl:
Rebased to v3.3.5 and addressed most of code review point that seem worth addressing. But I will not be able to find time to re-test this properly in the next few days, last changes are not yet validated.

src/mod/auth/auth.go
```diff
@@ -344,142 +599,1147 @@ func (a *AuthAgent) CheckAuth(r *http.Request) bool {
...
+
+	var assertPayload map[string]any
+	if err := json.NewDecoder(r.Body).Decode(&assertPayload); err != nil {
+		utils.SendErrorResponse(w, "failed to parse assertion data")
+		return
+	}
+
+	// Base64 trickery in the decodeBase64url is due to how browser part encodes the payload
+	// Error are omitted here as the actual contents will be validated via protocol.ParseCredentialRequestResponse
+	resp := assertPayload["response"].(map[string]any)
```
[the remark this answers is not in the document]
mvasl: Proposed fix will not handle unsanitized JSON either and the worst that could happen is `protocol.ParseCredentialRequestResponse` will choke on malformed data and return an error, failing the request.

src/mod/auth/auth.go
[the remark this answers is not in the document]
tobychui: There should be a nightly task ticker somewhere that is use for daily summary and statistics cleanup. But that is from another module, so using that for auth database cleaning might not be a good idea from the scope of clean code and oop design.

I guess you can add a new nightly ticker to clean that up in the auth module if needed.

tobychui:
Looking good, let me do a "human review" later this weekend.

src/mod/auth/auth.go
```diff
@@ -38,6 +70,207 @@ type AuthEndpoints struct {
...
+
+// deserializeSessionData converts JSON bytes back to *webauthnlib.SessionData.
+func deserializeSessionData(data []byte) (*webauthnlib.SessionData, error) {
+	var wrapper sessionDataWrapper
+	if err := json.Unmarshal(data, &wrapper); err != nil {
+		return nil, err
+	}
+
+	// Restore UserVerificationRequirement from string
+	var userVer protocol.UserVerificationRequirement
```
tobychui, line 173: enum (type & const in golang) should be used here for better future maintainability, but these shd be a simple fix so I will do it later.

src/mod/auth/auth.go
```diff
@@ -38,6 +70,207 @@ type AuthEndpoints struct {
...
+	case "preferred":
+		userVer = protocol.VerificationPreferred
+	case "discouraged":
+		userVer = protocol.VerificationDiscouraged
+	default:
+		userVer = protocol.VerificationPreferred
+	}
+
+	// Restore CredentialMediationRequirement from string
+	var mediation protocol.CredentialMediationRequirement
```
tobychui, line 186: same, enum shd be used here

src/mod/auth/auth.go
```diff
@@ -38,6 +70,207 @@ type AuthEndpoints struct {
...
+	default:
+		mediation = protocol.MediationDefault
+	}
+
+	// Restore CredParams
+	credParams := make([]protocol.CredentialParameter, len(wrapper.CredParams))
+	for i, cp := range wrapper.CredParams {
+		alg, _ := cp["alg"].(float64)
+		credParams[i] = protocol.CredentialParameter{
+			Type:      protocol.CredentialType(cp["type"].(string)),
```
tobychui, line 203: You didn't catch type assertion error here, might cause program panic

src/mod/auth/auth.go
```diff
@@ -38,6 +70,207 @@ type AuthEndpoints struct {
...
+
+// newWebAuthnFromRequest creates a WebAuthn instance configured for the origin of the current request.
+func newWebAuthnFromRequest(r *http.Request) (*webauthnlib.WebAuthn, error) {
+	host := r.Host
+	if fwdHost := r.Header.Get("X-Forwarded-Host"); fwdHost != "" {
+		host = strings.Split(fwdHost, ",")[0]
+		host = strings.TrimSpace(host)
+	}
+
+	// Strip port for RPID (bracket-safe for IPv6)
```
tobychui, line 255: Don't do it manually, use `net.SplitHostPort`

src/mod/auth/auth.go
```diff
@@ -54,6 +287,15 @@ func NewAuthenticationAgent(sessionName string, key []byte, sysdb *db.Database,
 		Database:                sysdb,
 		LoginRedirectionHandler: loginRedirectionHandler,
 		Logger:                  systemLogger,
+
+		totpMaxFailures:     5,
```
tobychui, line 291: These values should not be hard coded here. Please make them pass-in via function parameters that is set in the main.go scope. I will add start flags for them so user can adjust them via start flags.

src/mod/auth/auth.go
```diff
@@ -38,6 +70,207 @@ type AuthEndpoints struct {
...
+		mediation = protocol.MediationDefault
+	}
+
+	// Restore CredParams
+	credParams := make([]protocol.CredentialParameter, len(wrapper.CredParams))
+	for i, cp := range wrapper.CredParams {
+		alg, _ := cp["alg"].(float64)
+		credParams[i] = protocol.CredentialParameter{
+			Type:      protocol.CredentialType(cp["type"].(string)),
+			Algorithm: webauthncose.COSEAlgorithmIdentifier(alg),
```
tobychui, line 204: Same type assertion issue here,

src/mod/auth/auth.go
```diff
@@ -122,10 +364,23 @@ func (a *AuthAgent) HandleLogin(w http.ResponseWriter, r *http.Request) {
...
-		// Set user as authenticated
-		a.LoginUserByRequest(w, r, username, rememberme)
+		if a.IsTOTPEnabled(username) {
+			// Set session data but keep authenticated=false while waiting for TOTP
+			session := a.PrepareUserSession(w, r, username, rememberme)
+			session.Values["totp_partial"] = true
+			session.Values["authenticated"] = false
+			session.Save(r, w)
+
+			utils.SendJSONResponse(w, `{"totp_required":true,"username":"`+username+`"}`)
```
tobychui, line 374: Please use a function scoped struct for better maintainability.

src/mod/auth/auth.go
```diff
@@ -344,9 +599,376 @@ func (a *AuthAgent) CheckAuth(r *http.Request) bool {
...
+		return false
+	}
+
+	valid := totp.Validate(code, secret)
+	return valid
+}
+
+// VerifyTOTPCodeAndEnable verifies a TOTP code and enables 2FA for the user
+func (a *AuthAgent) VerifyTOTPCodeAndEnable(username, code string) (bool, error) {
+	var secret string
```
tobychui, line 689: Isn't this part of the flow just duplicates of `func (a *AuthAgent) VerifyTOTPCode(username, code string) bool`? You could reuse `VerifyTOTPCode()` instead.

src/mod/auth/auth.go
```diff
@@ -344,9 +599,376 @@ func (a *AuthAgent) CheckAuth(r *http.Request) bool {
...
+// UpdateTOTPSession marks the session as fully authenticated after TOTP verification
+func (a *AuthAgent) UpdateTOTPSession(w http.ResponseWriter, r *http.Request, username string) error {
+	session, err := a.SessionStore.Get(r, a.SessionName)
+	if err != nil {
+		return err
+	}
+
+	session.Values["authenticated"] = true
+	session.Values["username"] = username
+
```
tobychui, line 739: Just wondering do you need to unset the [totp_partial] session key here?

src/mod/auth/auth.go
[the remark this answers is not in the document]
tobychui: Copilot recommended fix is wrong but the issue is real. According to best OOP practice you should provide a `NewWebAuthnUser()` constructor for this and let the constructor populate the aAgent here. Directly creating a struct here is not maintainable in long term, one breaking change to the struct structure will require a large rewrite.

src/mod/auth/auth.go
```diff
@@ -483,3 +1119,597 @@ func Hash(raw string) string {
...
+		return
+	}
+	a.Database.Write("auth", fmt.Sprintf("webauthn/pending_reg/%s", token), map[string]any{
+		"username":    username,
+		"sessionData": sessionJSON,
+		"expiry":      time.Now().Add(5 * time.Minute).Format(time.RFC3339),
+	})
+
+	w.Header().Set("Content-Type", "application/json")
+	json.NewEncoder(w).Encode(map[string]any{
```
tobychui, line 1162: Please create a function scoped struct instead of using map[string]any (or interface{})

src/mod/auth/auth.go
```diff
@@ -483,3 +1119,597 @@ func Hash(raw string) string {
...
+		return
+	}
+	sessionData, err := deserializeSessionData(sessionDataBytes)
+	if err != nil {
+		a.Logger.PrintAndLog("auth", "failed to deserialize webauthn session data", err)
+		utils.SendErrorResponse(w, "session data parse error: "+err.Error())
+		return
+	}
+
+	user := &webAuthnUser{Username: username}
```
tobychui, line 1226: Same here, please replace this with the new webAuthnUser Constructor mentioned earlier for proper user initialization.

src/mod/auth/auth.go
```diff
@@ -483,3 +1119,597 @@ func Hash(raw string) string {
...
+		return
+	}
+
+	// Decode base64url fields to raw bytes
+	decodeBase64url := func(val any) ([]byte, error) {
+		s, ok := val.(string)
+		if !ok {
+			return nil, fmt.Errorf("expected string, got %T", val)
+		}
+		// Replace URL-safe characters with standard base64
```
tobychui, line 1242: Have you tried using `base64.RawURLEncoding.DecodeString(s)` instead of manually filtering out the URL padded characters?

src/mod/auth/auth.go
```diff
@@ -483,3 +1119,597 @@ func Hash(raw string) string {
...
+		}
+		return base64.StdEncoding.DecodeString(s)
+	}
+
+	rawId, err := decodeBase64url(credPayload["rawId"])
+	if err != nil {
+		utils.SendErrorResponse(w, "invalid rawId in credential")
+		return
+	}
+	clientDataJSON, err := decodeBase64url(credPayload["response"].(map[string]any)["clientDataJSON"])
```
tobychui, line 1257: Please check `credPayload["response"]` is not nil before type assertion, you also need to catch error for type assertion fail

src/mod/auth/auth.go
```diff
@@ -483,3 +1119,597 @@ func Hash(raw string) string {
...
+	if err != nil {
+		utils.SendErrorResponse(w, "invalid clientDataJSON")
+		return
+	}
+	attestationObject, err := decodeBase64url(credPayload["response"].(map[string]any)["attestationObject"])
```
tobychui, line 1262: same type check and assert error catch is needed here

src/mod/auth/auth.go
```diff
@@ -483,3 +1119,597 @@ func Hash(raw string) string {
...
+		utils.SendErrorResponse(w, "failed to prepare auth: "+err.Error())
+		return
+	}
+	a.Database.Write("auth", fmt.Sprintf("webauthn/pending_auth/%s", token), map[string]any{
+		"sessionData": sessionJSON,
+		"expiry":      time.Now().Add(5 * time.Minute).Format(time.RFC3339),
+	})
+
+	w.Header().Set("Content-Type", "application/json")
+	json.NewEncoder(w).Encode(map[string]any{
```
tobychui, line 1505: I think I kinda see this structure quite a few times in this module. Maybe you can make this struct a global struct in this module so it can be reused  when returning from REST endpoints?

src/mod/auth/auth.go
```diff
@@ -483,3 +1119,597 @@ func Hash(raw string) string {
...
+
+// loadPendingAuthSession loads and validates a pending auth session from the database.
+// Returns nil (without error) when the token is missing or expired.
+func (a *AuthAgent) loadPendingAuthSession(token string) (*pendingAuthSession, error) {
+	var pendingData map[string]any
+	if err := a.Database.Read("auth", fmt.Sprintf("webauthn/pending_auth/%s", token), &pendingData); err != nil {
+		return nil, err
+	}
+
+	expiryStr, ok := pendingData["expiry"].(string)
```
tobychui, line 1525: Check `pendingData["expiry"]` not nil before assert to prevent panic

src/mod/auth/auth.go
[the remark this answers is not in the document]
tobychui: No if you don't catch it it will panic (crash). Either way you need the ok check here, this is just how golang works.

src/mod/auth/auth.go
```diff
@@ -483,3 +1119,597 @@ func Hash(raw string) string {
...
+
+	// Finally, validate the passkey and user data
+	foundUsername := ""
+	_, credential, err := wa.ValidatePasskeyLogin(func(rawID, userHandle []byte) (webauthnlib.User, error) {
+		if len(userHandle) > 0 {
+			foundUsername = string(userHandle)
+			if !a.UserExists(foundUsername) {
+				return nil, errors.New("user not found")
+			}
+			return &webAuthnUser{aAgent: a, Username: foundUsername}, nil
```
tobychui, line 1642: Please replace this with a proper constructor

src/mod/auth/auth.go
```diff
@@ -483,3 +1119,597 @@ func Hash(raw string) string {
...
+	if err := sess.Save(r, w); err != nil {
+		utils.SendErrorResponse(w, "Failed to complete authentication")
+		return
+	}
+
+	a.Logger.PrintAndLog("auth", fmt.Sprintf("User %s logged in via WebAuthn passkey", foundUsername), nil)
+	utils.SendJSONResponse(w, `{"success":true,"username":"`+foundUsername+`"}`)
+}
+
+func (a *AuthAgent) decodeBase64url(val any) ([]byte, error) {
```
tobychui, line 1688: I wonder won't this just works without those ReplaceAll parts? 
```
func (a *AuthAgent) decodeBase64url(val any) ([]byte, error) {
	s, ok := val.(string)
	if !ok || s == "" {
		return []byte{}, nil
	}
	return base64.URLEncoding.WithPadding(base64.NoPadding).DecodeString(s)
}
```

tobychui (changes requested):
The authentication flow looks good to me, a few things require changes

- Some of the code logic are duplicated, and some are already covered by latest golang packages as build in functions so you no longer need to manually write them here. Please update them according to the comments.
- A few places you forgot to check if map field is null and proceed directly to type assertion. This will definitely cause panic (crash) which is not ideal for system stability, please try to fix them in this revision. 
- The rate limit with proxy in front situations, I think later on we can add a check to see if the X-Forward-From latest ip is from a trusted proxy, if yes we trim it off and use the 2nd latest ip for rate limit check (or continue checking until the one that is not in our trusted proxy list with up to n ips)
- I didn't review the front-end, since it should not be related to security nor stability, if it works then don't touch it.

I understand there is no true OOP in golang, but it is still a good practice to use OOP like constructor and classes access function in coding Go codebase. The rest of the backend code looks fine, and multi-user support part is optional so you can ignore that change request from copilot for now.

src/mod/auth/auth.go
mvasl: This is just a shortcut to keep lines shorter, but i agree that sometimes it might hurt readability in long functions

src/mod/auth/auth.go
mvasl: Per 5.3 of webauth spec these fields are not optional and must exist
https://w3c.github.io/webauthn/#dictdef-publickeycredentialparameters

tobychui: Yes I understand specification requires it and in normal usecase it will not be nil, but you still need to guard it in case attacker tries to crash your server with custom engineered payload, then you will see a CVE which highlight this line.

PR #1260, opened 2026-08-09
plugin store: add opt-in preview flag to plugin introspect

reptil1990:
The app store derives the "Preview" tag solely from a missing icon.png, so a finished plugin without an icon is flagged as preview and a plugin cannot deliberately mark itself as preview.

Add an optional preview boolean to the plugin IntroSpect. The field is needed on the Go struct so the flag survives the JSON round-trip into the store frontend. The existing missing-icon fallback stays for backward compatibility.

### DIFF
--- a/docs/plugins/docs/2. Architecture/2. Introspect.md
+++ b/docs/plugins/docs/2. Architecture/2. Introspect.md
@@ -1,6 +1,6 @@
 # Introspect
 
-Last Update: 25/05/2025
+Last Update: 09/08/2026
 
 ---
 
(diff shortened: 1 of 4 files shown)


tobychui (approved):
LGTM, will be adding this in next rc-1 (release candidate specifically for new feature integrations)

reptil1990:
Tonight I will look into the plugin-repo to see if there are changes necessary but i think so.

AnthonyMichaelTDM (approved):
lgtm

AnthonyMichaelTDM:
> Tonight I will look into the plugin-repo to see if there are changes necessary but i think so.

Yeah you'll need to make the dirupdate2 tool in the plugin store repo fill out the field, and then regenerate the index

reptil1990:
> > Tonight I will look into the plugin-repo to see if there are changes necessary but i think so.
> 
> Yeah you'll need to make the dirupdate2 tool in the plugin store repo fill out the field, and then regenerate the index

Yeah I thought so.. it is already in:
[‎tools/dirupdate/mod/zoraxy_plugin/zoraxy_plugin.go](https://github.com/aroz-online/zoraxy-official-plugins/pull/19/changes#diff-b01066d0d435301194aa4bbc4851a312b5945d344f9219eafb4bf25bcc1d8c39)

PR #1261, opened 2026-08-12
fix(ui): reload access rule lists when the tab is opened

driin0:
### What happens

Open Access Control, switch to another tab, come back: the black/whitelist tables still show what they showed the first time. Only a full page reload updates them.

### Why

`initBlackWhitelistTables()` is called from two places in `access.html` — when the component first renders, and from `handleSelectedAccessRuleChange()`. Nothing runs it when the tab is re-opened, so anything that changed the lists in between is invisible.

This shows up whenever the lists are changed outside that page: a second admin session, or a plugin keeping a whitelist in sync. In my case the entries are IPs resolved from DDNS names, so they move on their own every few minutes and the page is almost always out of date.

### The change

`index.html` already provides `tabSwitchEventBind`, and `status`, `httprp`, `vdir`, `rules`, `setroot`, `streamproxy` and `pluginContextWindow` all register there. This adds `access` alongside them. No new mechanism.

### Tested

Zoraxy 3.3.3 in Docker (arm64) running with `-dev` against this branch's `src/web`: changed a whitelisted IP from outside the page, switched to Status and back, the table showed the new value without a reload. Verified the same sequence still shows the stale value on an unpatched build.

### DIFF
--- a/src/web/components/access.html
+++ b/src/web/components/access.html
@@ -1677,4 +1677,13 @@
     // Scope to this page's accordion only; .advanceSettings is reused by
     // other webmin components loaded dynamically via ajax.
     $("#accessAdvanceSettings").accordion();
+
+    //Bind on tab switch events
+    tabSwitchEventBind["access"] = function(){
+        //The lists are only loaded when this component is first rendered or
+        //when another access rule is picked. Anything that changed them in
+        //between - another session, or a plugin syncing IPs into a whitelist -
+        //stays invisible until a full page reload, so reload them here.
+        initBlackWhitelistTables();
+    };
 </script>


tobychui:
Hey @driin0 

Good patch, I have read the PR code comment regarding the change. 

First of all your plugin should not call API from the front-end, it should use the plugin REST API endpoint from backend. But this fix is valid in the case where someone opened multiple tabs for editing (state sync between tabs). I will merge this regardless. 

If your plugin(s) is depending on another front-end API (which I cannot guarantee it is always backward compatible), it would be better for you to switch over to the plugin rest api instead.

tobychui (approved):

driin0:
Thanks for merging, and for the pointer.

Small clarification on the case I described: the plugin I was referring to does
use the plugin REST API from its backend — it talks to /plugin/api/access/list and
/plugin/api/whitelist/ip/{add,remove} with the Bearer key, no front-end calls
to Zoraxy at all. The stale table was just what I was looking at while the
backend changed the list underneath.

The point you raise does apply to a second plugin of mine, though, and there
it isn't a choice: the endpoints it needs to write to aren't in plugin_api.go.
I've put the details and the questions they raise in a discussion, rather than
clutter a merged PR: https://github.com/tobychui/zoraxy/discussions/1269

PR #1263, opened 2026-08-14
Bind OAuth2 SSO sessions to tokens issued by Zoraxy

Elandrya2711:
The OAuth2 authorization check takes the `z-token` cookie value as a bearer token and calls the provider's UserInfo endpoint with it. Any HTTP 200 response is accepted as authorized. Nothing ties that token to the OAuth2 client Zoraxy is configured with, so a token issued by the same identity provider for a different client passes the check as well.

This matters as soon as the provider serves more than one application: a token that legitimately belongs to another client is accepted by Zoraxy. The raw token is also handed on to upstream applications together with the remaining cookies.

This change stores the token server side in a TTL cache and puts an opaque random session id in the cookie instead. Only tokens obtained through Zoraxy's own code exchange can be used, and the token no longer travels to upstream applications. The per-request UserInfo call is kept, so revoked tokens still end the session.

Behaviour changes worth considering before merging:

- SSO sessions are held in memory, so restarting Zoraxy signs users out.
- Upstream applications no longer receive the access token in `z-token`.
- Saving the OAuth2 settings flushes existing sessions.

An alternative approach would be to validate the `aud` claim of an ID token against the configured client id via JWKS. That keeps the check stateless, but it adds a JWT dependency and requires the provider to issue ID tokens. Happy to rework it that way if you prefer.

Builds on #1262.

### DIFF
--- a/src/mod/auth/sso/oauth2/oauth2.go
+++ b/src/mod/auth/sso/oauth2/oauth2.go
@@ -2,6 +2,8 @@ package oauth2
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -35,6 +37,7 @@ type OAuth2RouterOptions struct {
 	Logger                       *logger.Logger
 	Database                     *database.Database
 	OAuth2ConfigCache            *ttlcache.Cache[string, *oauth2.Config]
+	OAuth2SessionStore           *ttlcache.Cache[string, *oauth2.Token]
 }
 
 type OIDCDiscoveryDocument struct {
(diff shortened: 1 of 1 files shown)


tobychui:
The oauth (similar to forward auth), is a community maintained feature that I dont have infrastructure to test.
Maybe you could ping [@kjagosz](https://github.com/kjagosz), the author of the Oauth2 module, and see if he got any idea on this?

AnthonyMichaelTDM:
If the PRs are meant to be additive, could you arrange them in a stack?

https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests

Elandrya2711:
They are stacked in git already (oauth2-session-store = oauth2-httponly-cookies + one commit), GitHub just cannot show it since a fork PR cannot use a fork branch as base. Once #1262 is merged I rebase this one right after.

PR #1274, opened 2026-08-22
Update CHANGELOG.md

Morethanevil:
Is this file still needed? You have the changes already within the release notes


tobychui:
Yes, because the changes in the release log doesn't includes all actual changes. Sometime copilot kinda miss out a few features when drafting that. It is always better to have a human in the loop to make sure AI don't screw up :)

tobychui (approved):

PR #1292, opened 2026-09-13
Fix #1283

### LINKED ISSUE #1283: [ACME][BUG] DNS-01: propagation timeout is silently reset to 0, and the recursive propagation check is force-disabled
Two independent problems in `src/mod/acme/acme.go` that both make DNS-01 challenges fail in ways that look like "the TXT record is deleted too early". Found while debugging #351 (IPv64), but neither is provider-specific — both affect every DNS provider.

Versions: Zoraxy `main` (3.2.x), lego v5.

---

## Bug 1 — `PropTimeout` from the cert JSON overwrites the configured timeout, even when it is 0

`src/mod/acme/acme.go`, in `ObtainCert()`:

```go
certInfo, err := LoadCertInfoJSON(fmt.Sprintf("./conf/certs/%s.json", certificateName))
if err == nil {
    useDNS = certInfo.UseDNS
    if dnsServers == "" && certInfo.DNSServers != nil && len(certInfo.DNSServers) > 0 {
        dnsServers = strings.Join(certInfo.DNSServers, ",")
    }
    propagationTimeout = certInfo.PropTimeout   // <-- no > 0 check
}
```

`CertificateInfoJSON.PropTimeout` is `int` with `json:"prop_time"`. Any `conf/certs/<name>.json` that does not contain `prop_time` (certificates created by an older Zoraxy version, hand-written files, …) unmarshals to **0**, and that 0 unconditionally replaces the value that the handler carefully validated a few lines earlier:

```go
// Default propagation timeout is 600 seconds (10 minutes)
propagationTimeout := 600
...
if propagationTimeout < 60 {
    // Minimum propagation timeout is 60 seconds
    propagationTimeout = 60
}
```

The 0 then travels through `GetDnsChallengeProviderByName(dnsProvider, dnsCredentials, propagationTimeout)` → `acme_dns.go`'s `userDefinedPropagationTimeout := ppgTimeout` → `acmedns.GetDNSProviderByJsonConfig(...)` → the provider's `Config.PropagationTimeout` → `provider.Timeout()`.

In lego, `challenge/dns01/dns_challenge.go`:

```go
timeout, interval = provider.Timeout()
...
time.Sleep(interval)
err = wait.For(timeout, interval, func() (bool, error) { ... })
```

and `internal/wait/wait.go`:

```go
timeUp := time.After(timeout)
for {
    select {
    case <-timeUp:
        return errors.New("time limit exceeded")
    default:
    }
    ...
}
```

With `timeout == 0`, `time.After(0)` is ready immediately, so `wait.For` returns `time limit exceeded` on the very first iteration.

**Observable behaviour:** the TXT record is created, and roughly 2 seconds later (the single `time.Sleep(interval)`) lego's deferred cleanup deletes it again. From the DNS provider's side this looks exactly like "Zoraxy adds the record and instantly removes it". I believe this is also what #353 is reporting.

### Suggested fix

```go
if certInfo.PropTimeout > 0 {
    propagationTimeout = certInfo.PropTimeout
}
```

Optionally also clamp to the same 60 s minimum that the HTTP handler enforces, so a hand-edited JSON cannot go below it either.

---

## Bug 2 — `DisableRecursiveNSsPropagationRequirement()` is hard-coded

`src/mod/acme/acme.go`:

```go
opts := &dns01.Options{}
if len(dnsNameservers) > 0 && dnsNameservers[0] != "" {
    opts.RecursiveNameservers = dnsNameservers
} else {
    opts.RecursiveNameservers = defaultNameservers   // 8.8.8.8, 8.8.4.4, 1.1.1.1, 1.0.0.1
}
dns01.SetDefaultClient(dns01.NewClient(opts))

err = client.Challenge.SetDNS01Provider(provider, dns01.DisableRecursiveNSsPropagationRequirement())
```

lego v5 enables both propagation checks by default (`challenge/dns01/dns_challenge_precheck.go`):

```go
func newPreCheck() preCheck {
    return preCheck{
        requireAuthoritativeNssPropagation: true,
        requireRecursiveNssPropagation:     true,
    }
}
```

Zoraxy disables the recursive one for every certificate, with no way to turn it back on. The configured recursive nameservers are then still used for CNAME resolution and for locating the authoritative NS, but they no longer gate propagation — so the carefully configured "DNS servers" setting no longer does what a user would expect it to do.

**Why this matters now:** since 2025 Let's Encrypt performs Multi-Perspective Issuance Corroboration. The secondary perspectives resolve the challenge name through *their own* recursive resolvers. lego, checking only the zone's authoritative nameservers, reports "propagated" within ~2 seconds — while public resolvers may still be serving a cached negative answer for `_acme-challenge.<domain>`, whose lifetime is `min(SOA MINIMUM, SOA record TTL)` (RFC 2308 §4). Validation then fails with:

```
403 ... during secondary validation: DNS problem: no TXT record found for _acme-challenge.<domain>
```

Concrete example from #351: the IPv64 zones had an effective negative-caching TTL of 3600 s, so every failed attempt poisoned the next hour of retries. The authoritative-only check cannot see that at all. (IPv64 is lowering those SOA values, but the same situation arises with any provider whose zones have a long negative TTL.)

### Suggested fix

Do not hard-code the option. Either drop it and use lego's defaults, or expose it per certificate — same place the propagation timeout and DNS servers are already configurable. `dns01.PropagationWait(d, skipCheck)` is another option lego offers if the goal was to avoid slow authoritative checks.

---

## Not a Zoraxy bug, for context

The original failure in #351 is upstream in lego: `providers/dns/ipv64/ipv64.go`'s `splitDomain()` assumes the last three labels are always the zone, which only holds for `<host>.ipv64.net` and breaks for every custom domain. IPv64 has deployed a server-side normalisation that absorbs the malformed requests, so #351 should stop reproducing without any Zoraxy change — but a lego-side fix is still warranted. I will open that separately at go-acme/lego.

zen8841:
Fix #1283

add check for propgation timeout value

add switch to enable recursive ns propagation check

TODO: Add a GUI option to switch recursive NS propagation checks (currently, this can only be set in domain.json).


tobychui (approved):
LGTM

Issue #1301, opened 2026-09-17
[ACME] DNS-01: Add DisableAuthoritativeNssPropagationRequirement option


PetziAt:
**Describe the bug**
When using the ACME DNS-01 challenge with the OVH DNS provider, Zoraxy uses the configured internal DNS resolver for zone discovery, but Lego subsequently tries to query the authoritative DNS nameservers directly over UDP/53.

Direct outbound DNS traffic to Internet DNS servers is intentionally blocked in this environment. DNS resolution is handled exclusively through an internal Unbound resolver.

The configured DNS resolver works correctly. The ACME request fails because Lego attempts to contact the authoritative OVH nameserver directly instead of using the configured recursive resolver for the propagation check.

**Request Details**
- ACME mode: DNS challenge
- CA used: Let's Encrypt
- DNS provider: OVH

**Error Message**
Copy and paste the whole error message here with code block tag (i.e. 3 x "`" before and after the code area).

```text
resolver: one or more domains had a problem: [host.subdomain.example.com: dns01: time limit exceeded: last error: authoritative nameservers: DNS call error: read udp 10.0.0.10:xxxxx->203.0.113.10:53: read: no route to host [ns=ns.example-dns.com.:53, question='_acme-challenge.subdomain.example.com. IN TXT']]
```

**To Reproduce**
1. Configure Zoraxy ACME with DNS challenge.
2. Select the OVH DNS provider.
3. Configure an internal recursive DNS resolver.
4. Block direct outbound DNS traffic to Internet nameservers on UDP/53.
5. Request a certificate for `host.subdomain.example.com`.
6. Zoraxy can resolve the DNS zone successfully through the configured internal resolver.
7. During the DNS-01 propagation check, Lego attempts to query the authoritative OVH nameserver directly.
8. The request eventually fails with `no route to host` / timeout.

**Expected Common Name (CN) in certy**
[
  "host.subdomain.example.com"
]

**Additional context**
The internal recursive DNS resolver is working correctly and can resolve the authoritative SOA for the domain.

The OVH DNS API does not require direct DNS access from Zoraxy. The OVH provider can create the `_acme-challenge` TXT record through the OVH API.

The failure occurs during the subsequent DNS propagation check, before the certificate can be issued.

Direct outbound DNS traffic from the Zoraxy host/LXC to Internet DNS servers is intentionally blocked. All DNS resolution is handled through the internal Unbound resolver.

[...]

PR #1303, opened 2026-09-18
Add granular DNS-01 NS propagation check controls

zen8841:
## Summary

add feature for #1301 

Replace the single "Enable Recursive NS Propagation Check" toggle with two
independent disable flags, giving users finer-grained control over DNS-01
propagation verification.

### Backend (`acme.go`)
- Rename `UseRecursiveNS` → `DisableRecursiveNssCheck` (inverted semantics)
- Add `DisableAuthoritativeNssCheck` field to `CertificateInfoJSON`
- Update `ObtainCert` signature to accept both disable flags
- Add guard: disabling authoritative NS check forces recursive NS check to
  stay enabled, ensuring at least one propagation check is always active
- Pass both flags to lego's `dns01.ChallengeOption`

### UI (`acme.html`)
- Replace single checkbox with two independent disable checkboxes
- Add mutual exclusion logic: checking one unchecks the other
- Default state: only "Disable Recursive NSs" is checked
- Update help text to document both checks and their behavior

### Migration (`v335/v335.go`, `updatelogic.go`)
- Add v3.3.4→v3.3.5 migration routine that transforms existing cert JSON:
  - `recursive_ns: true` → `disable_recursive_nss_check: false`
  - `recursive_ns: false` → `disable_recursive_nss_check: true`
  - Adds `disable_authoritative_nss_check: false` where missing
- Skip files already migrated or missing the old key


src/mod/acme/autorenew.go
```diff
@@ -406,7 +406,7 @@ func (a *AutoRenewer) renewExpiredDomains(certs []*ExpiredCerts) ([]string, erro
 			a.Logf("Could not extract SANs from PEM for "+fileName+", using original domains", errSan)
 		}
 
-		_, err = a.AcmeHandler.ObtainCert(expiredCert.Domains, certName, a.RenewerConfig.Email, certInfo.AcmeName, certInfo.AcmeUrl, certInfo.SkipTLS, certInfo.UseDNS, certInfo.PropTimeout, dnsServers, certInfo.UseRecursiveNS)
+		_, err = a.AcmeHandler.ObtainCert(expiredCert.Domains, certName, a.RenewerConfig.Email, certInfo.AcmeName, certInfo.AcmeUrl, certInfo.SkipTLS, certInfo.UseDNS, certInfo.PropTimeout, dnsServers, certInfo.DisableRecursiveNssCheck, certInfo.DisableAuthoritativeNssCheck)
```
tobychui, line 409: I think later we should make ObtainCert pass via a custom struct (e.g. `ObtainCertConfig{}`) currently the intake parameters are getting a bit too long.

src/mod/update/v335/v335.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// UpdateFrom334To335 migrates cert JSON files to the new NSs check field names.
+func UpdateFrom334To335() error {
```
tobychui, line 17: This update is a bit tricky here, since we now have rc-1 released, this change might cause some problem if user is upgrading from v3.3.5-rc1 to v3.3.5-rc2. 

Please also add a probe function that check if current version == v3.3.5 && struct not yet updated => if yes, run an update again even if current version is already v3.3.5.

src/mod/update/updatelogic.go
```diff
@@ -35,6 +36,12 @@ func runUpdateRoutineWithVersion(fromVersion int, toVersion int) {
 		if err := v334.UpdateFrom333To334(); err != nil {
 			fmt.Println("Warning: fallback cert migration failed (non-fatal):", err)
 		}
+	} else if fromVersion == 334 && toVersion == 335 {
```
tobychui, line 39: This need to be changed to `else if (fromVersion == 334 && toVersion == 335) || (fromVersion  == 335 && probe_config())` as mentioned above

tobychui (commented):
The new config update logic might cause some issue with our current release schedule. I guess a new probe function is needed to make sure the update don't crash out the users who are using the v3.3.5-rc1 release.

src/mod/acme/autorenew.go
zen8841: You and I have the same idea; I also plan to do this. The Certinfo struct actually already contains most of the information.
I'm also considering whether to migrate the data from domain.json to the database, and then provide an interface to change the stored cert issue settings.

tobychui (approved):
LGTM

PR #1308, opened 2026-09-21
perf(web): enable revalidatable caching for static assets and snippet HTML

yw9381:
## Summary

Static resources are served from the embedded filesystem, whose files report a zero modtime. `http.FileServer` therefore emits no `Last-Modified`/`ETag` and no `Cache-Control`, so browsers have no validator and cannot reuse any cached copy. As a result `jquery`, `semantic` and `utils.js` were re-downloaded on every iframe load, and the `?t=Date.now()` cache busters on snippet URLs made the snippet HTML uncacheable as well.

## Changes

- Serve static resources with a content-based strong `ETag` plus `Cache-Control: no-cache`, so browsers revalidate and reuse the cached body (`304`) instead of re-downloading it. Hashes are cached per path in production and recomputed in development to keep hot reload working.
- Emit an `ETag` over the rendered HTML after CSRF token injection. The validator is session-specific but still allows `304` revalidation.
- Add `setIframeSrcWithReload()` to force a real iframe navigation when the URL only differs by the `#payload` fragment, and use it in `showSideWrapper` and `showEditorSideWrapper` so snippet inline scripts always re-run with fresh data.
- Drop the `?t=Date.now()` cache busters, now handled by server-side revalidation.

## Notes

No behavior change to routing or authentication; only response caching headers and iframe loading logic.


tobychui:
No because this implementation breaks how dev flag should works, when in development build (i.e. start with `-dev` flag, for UI devs working with UI updates) the admin UI should be reloaded every time from backend. 

Although I agree that non-dev mode caching is one of the optimization that can be made, this PR breaks the dev mode behavior so I cannot merge this.

PR #1309, opened 2026-09-21
fix: correct OAuth2 User Info URL field name casing

Samet195:
The User Info URL input was named `oauth2UserInfoURL` while the backend reads the POST parameter as `oauth2UserInfoUrl`. Form keys are matched case-sensitively, so saving the OAuth2 settings in manual mode (without a discovery URL) failed with "oauth2UserInfoUrl not found". The id mismatch also prevented the field from being populated on page load.

Rename the label target, input id/name and the JS selector to `oauth2UserInfoUrl` so they match the backend and each other.


tobychui (approved):
LGTM

